PR #40, opened 2019-04-01
Implement observers for ETH and XRP

leoneparise:
This pull request contains a first implementation to register transaction observers on new blocks for Ethereum and Ripple.


terorie:
Included in #126 instead

PR #46, opened 2019-04-09
Add Aion platform

kolya182:
Fixes https://github.com/TrustWallet/blockatlas/issues/32 https://github.com/TrustWallet/private-trust-api/issues/266


coin/slip44.go
```diff
@@ -8,14 +8,16 @@ type Coin struct {
 	Decimals uint   `json:"decimals"`
 }
 
+// List of coin indexes
```
vikmeup: no need comment

go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/trustwallet/blockatlas
 go 1.12
 
 require (
+	github.com/davecgh/go-spew v1.1.1
```
vikmeup: `go-spew` not used?

platform/aion/api.go
```diff
@@ -0,0 +1,58 @@
...
+	})
+	router.GET("/:address", getTransactions)
+}
+
+func getTransactions(c *gin.Context) {
+	trxs, err := client.GetTxsOfAddress(c.Param("address"))
+
+	if err != nil {
+		logrus.WithError(err).Errorf("Aion: Failed to get transactions for address %v", c.Param("address"))
+	}
```
vikmeup: add integration tests

go.mod
terorie: It's used in the tests. I don't like `go.mod` / `go.sum` it pollutes the commits. Maybe we should remove it.
@kolya182 You can use `export GO111MODULE=off` to disable go.mod if you want

coin/slip44.go
kolya182: Actually `go` recommends adding comments` https://blog.golang.org/godoc-documenting-go-code, plus linter was demanding adding it.

kolya182: many open-source `go` projects do comments https://github.com/cosmos/cosmos-sdk/blob/develop/server/init.go https://github.com/trezor/blockbook/blob/master/db/sync.go

terorie: Projects should follow godoc standard godoc.org/github.com/trustwallet/blockatlas
That’s why the linter was asking for it.
Feel free to add it but I’m going to document the whole source anyways once the code structure is set in stone.

loaders.go
```diff
@@ -10,6 +10,7 @@ import (
 	"github.com/trustwallet/blockatlas/platform/ripple"
 	"github.com/trustwallet/blockatlas/platform/stellar"
 	"github.com/trustwallet/blockatlas/platform/tezos"
+	"github.com/trustwallet/blockatlas/platform/aion"
```
terorie: Please use alphabetic sort for import (lint)

platform/aion/source/model.go
```diff
@@ -0,0 +1,18 @@
+package source
+
+type AionTransactionRPCResponse struct {
+	Content []AionTransaction 
+} 
+
+type AionTransaction struct {
```
terorie: No need for `Aion` prefix. The prefix is the package name already (`source.Transaction` vs `model.Transaction`)

platform/aion/source/model.go
```diff
@@ -0,0 +1,18 @@
+package source
+
+type AionTransactionRPCResponse struct {
```
terorie: How about `type TransactionPage struct` instead?

platform/aion/api.go
terorie: Yes, please add an Aion address to the map in the integration test file https://github.com/TrustWallet/blockatlas/blob/master/test/main.go

coin/slip44.go
terorie: Comments will be added in #50

PR #60, opened 2019-04-12
platform/ethereum forks (second try)

terorie:
Same as #59, but redone because GitHub is buggy


loaders.go
```diff
@@ -15,14 +15,21 @@ import (
...
-	"ethereum": ethereum.Setup,
-	"aion": 	aion.Setup,
+	"binance":          binance.Setup,
+	"nimiq":            nimiq.Setup,
+	"ripple":           ripple.Setup,
+	"stellar":          stellar.MakeSetup(coin.XLM, "stellar"),
+	"kin":              stellar.MakeSetup(coin.KIN, "kin"),
+	"tezos":            tezos.Setup,
+	"ethereum":         ethereum.MakeSetup(coin.ETH,  "ethereum"),
+	"ethereum-classic": ethereum.MakeSetup(coin.ETC,  "ethereum-classic"),
```
vikmeup: @terorie `ethereum-classic ` => `classic` can you change to this? for consistency

vikmeup: also `v1/ethereum-classic/0x4f9254c83eb525f9fcf346490bbb3ed28a81c661` doesn't seem to be working

terorie: Yes, will do

PR #78, opened 2019-04-22
platform/tron: Fix response

### LINKED ISSUE #77: TRON: wrong address, timestamp, and missing block returned
https://blockatlas.trustwalletapp.com/v1/tron/TKLnCNY5EsLNCvCXQTCn1dtqvc6vHhJUyJ

TODO:
- [x] Fix timestamp: 1555802447872 / 1000
- [x] Missing `block` property in response
- [x] `from` and `to` must be Tron address, not hex
- [x] Add test covers check for mentioned issues above

terorie:
Closes #77


platform/tron/api.go
```diff
@@ -38,13 +38,23 @@ func Normalize(srcTx *Tx) (tx models.Tx, ok bool) {
 	switch contract.Parameter.(type) {
 	case TransferContract:
 		transfer := contract.Parameter.(TransferContract)
+
+		from, err := HexToAddress(transfer.Value.OwnerAddress)
```
vikmeup: I would rather call this as `HexToBase58`

util/base58/base58.go
```diff
@@ -0,0 +1,172 @@
...
+)
+
+// Alphabet: copy from https://en.wikipedia.org/wiki/Base58
+var (
+	//BitcoinAlphabet = NewAlphabet("123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz")
+	//IPFSAlphabet    = NewAlphabet("123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz")
+	//FlickrAlphabet  = NewAlphabet("123456789abcdefghijkmnopqrstuvwxyzABCDEFGHJKLMNPQRSTUVWXYZ")
+	//RippleAlphabet  = NewAlphabet("rpshnaf39wBUDNEGHJKLM4PQRST7VWXYZ2bcdeCg65jkm8oFqi1tuvAxyz")
+	TronAlphabet = NewAlphabet("123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz")
+)
```
vikmeup: can you just bring `base58`? not sure why we need to have our own implementation

platform/tron/api.go
```diff
@@ -38,13 +38,23 @@ func Normalize(srcTx *Tx) (tx models.Tx, ok bool) {
...
+			ID:   srcTx.ID,
 			Coin: coin.TRX,
-			Date: srcTx.Data.Timestamp,
-			From: transfer.Value.OwnerAddress,
-			To: transfer.Value.ToAddress,
-			Fee: "0",
+			Date: srcTx.Data.Timestamp / 1000,
+			From: from,
+			To:   to,
+			Fee:  "0",
```
vikmeup: is there a way to know the fee?

util/base58/base58.go
terorie: Okay, I'll import the package instead.

platform/tron/api.go
terorie: How about `HexToBase58Check`? It isn't just standard base58 because it also contains a checksum.

vikmeup: `HexToBase58Check` make sense, I think in it's commonly used as base58 or base58 address

Issue #86, opened 2019-04-26
Wrong from direction returned


kolya182:
`Request: `https://blockatlas.trustwalletapp.com/v1/ethereum/0x9dfaC9bbe1119Da5DcdD7b095DB40920CDE878Ef

`Expected`: Trx with id `0xbb085371731a8990c12d87ea50252b04f552b7a34e6fbef4542b47b0d362b1d7` to have `to`:`0x4f833a24e1f95d70f028921e27040ca56e09ab0b` at the top level

`Actual`
`to`:`0x61b9898c9b60a159fc91ae8026563cd226b7a0c1`

Trust-Ray has correct response `from`:`0x61b9898c9b60a159fc91ae8026563cd226b7a0c1`
https://api.trustwallet.com/ethereum/transactions?address=0x9dfaC9bbe1119Da5DcdD7b095DB40920CDE878Ef

Etherescan ref
https://etherscan.io/tx/0xbb085371731a8990c12d87ea50252b04f552b7a34e6fbef4542b47b0d362b1d7

PR #88, opened 2019-04-28
Fix wrong recepient address in top level

kolya182:
Fixes https://github.com/TrustWallet/blockatlas/issues/86

### DIFF
--- a/platform/ethereum/api.go
+++ b/platform/ethereum/api.go
@@ -121,7 +121,7 @@ func AppendTxs(in []models.Tx, srcTx *Doc, coinIndex uint) (out []models.Tx) {
 	}
 
 	for _, op := range srcTx.Ops {
-		if op.Type == models.TxTokenTransfer && (op.From == baseTx.From || op.To == baseTx.To) {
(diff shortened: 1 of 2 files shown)


platform/ethereum/api.go
```diff
@@ -115,27 +115,26 @@ func AppendTxs(in []models.Tx, srcTx *Doc, coinIndex uint) (out []models.Tx) {
...
-			Name:     op.Contract.Name,
-			Symbol:   op.Contract.Symbol,
-			TokenID:  op.Contract.Address,
-			Decimals: op.Contract.Decimals,
-			Value:    models.Amount(op.Value),
-			From:     op.From,
-			To:       op.To,
+
+	for _, op := range srcTx.Ops {
+		if op.Type == models.TxTokenTransfer && (op.From == baseTx.From || op.To == baseTx.From) {
```
vikmeup: trust ray should only return 1 transaction that you need to use for token transfer, there is no need to loop. I think you need to leave the previous code as is and fix this on trust-ray

PR #90, opened 2019-04-29
Fix Kin missing transactions

kolya182:


platform/stellar/client.go
```diff
@@ -13,7 +13,7 @@ type Client struct {
 }
 
 func (c *Client) GetTxsOfAddress(address string) (txs []Payment, err error) {
-	path := fmt.Sprintf("%s/accounts/%s/payments?descending=true",
+	path := fmt.Sprintf("%s/accounts/%s/payments?order=desc&limit=25",
```
terorie: Using `models.TxPerPage` instead hardcoding it is recommended https://github.com/TrustWallet/blockatlas/blob/df5900e09586201b13b6525bdf909bbd121dd467/models/api.go#L23

PR #105, opened 2019-05-06
Add cosmos support

jamesscaur:
I wasn't able to figure out completely how the model, client, and api.go modules interact, but hopefully this is enough headway that it'll be easy to finish off... thanks for the opportunity to learn some Go.


platform/cosmos/model.go
```diff
@@ -0,0 +1,32 @@
...
+
+type Tx struct {
+	BlockHeight uint64 `json:"height"`
+	Timestamp   string `json:"timestamp"`
+	Hash        string `json:"txhash"`
+	TxData      struct {
+		TxType  string `json:"type"`
+		TxValue struct {
+			TxMessage struct {
+				TxContents []TxContents `json:"value"`
```
vikmeup: does not look good this way, can you split into separate models in the file?

terorie: I'll finish up the PR, how urgent is it?

jamesscaur:
Making a new one

PR #110, opened 2019-05-13
IOTX integration

dustinxie:
Closes https://github.com/TrustWallet/blockatlas/issues/101

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,6 +14,7 @@ require (
 	github.com/spf13/cast v1.3.0
 	github.com/spf13/cobra v0.0.3
 	github.com/spf13/viper v1.3.2
+	github.com/stretchr/testify v1.2.2
 	github.com/valyala/fastjson v1.4.1
 	github.com/ybbus/jsonrpc v2.1.2+incompatible
 	golang.org/x/net v0.0.0-20190514140710-3ec191127204 // indirect
--- a/platform/iotex/api.go
+++ b/platform/iotex/api.go
@@ -1,0 +1,110 @@
+package iotex
+
+import(
+	"net/http"
+	"strconv"
+	"time"
+
+	"github.com/gin-gonic/gin"
+	"github.com/spf13/viper"
+	"github.com/iotexproject/go-pkgs/crypto"
+	"github.com/iotexproject/iotex-address/address"
+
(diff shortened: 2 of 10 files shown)


terorie:
Thanks for your contribution! Is it possible to add support without introducing a dependency to `github.com/iotexproject`?
We try to keep coin-specific dependencies at a minimum for performance and security reasons (e.g. import times and package side-effects).

dustinxie:
> Thanks for your contribution! Is it possible to add support without introducing a dependency to `github.com/iotexproject`?
> We try to keep coin-specific dependencies at a minimum for performance and security reasons (e.g. import times and package side-effects).

Removed in latest push, pls take a look, thx!

platform/iotex/api.go
```diff
@@ -7,8 +7,6 @@ import(
 
 	"github.com/gin-gonic/gin"
 	"github.com/spf13/viper"
-	"github.com/iotexproject/go-pkgs/crypto"
-	"github.com/iotexproject/iotex-address/address"
```
dustinxie: removed

go.mod
```diff
@@ -3,21 +3,31 @@ module github.com/trustwallet/blockatlas
 go 1.12
 
 require (
-	github.com/davecgh/go-spew v1.1.1
+	github.com/BurntSushi/toml v0.3.1 // indirect
 	github.com/fsnotify/fsnotify v1.4.7
```
dustinxie: result of running 'go mod tidy'

dustinxie:
@terorie 
can you please take a look of new update/commit

platform/iotex/model.go
```diff
@@ -0,0 +1,39 @@
+package iotex
+
+const TransferFee = "10000000000000000"
```
vikmeup: is it possible to get transfer fee from the transaction object and not to hardcode?

dustinxie: our transfer fee is fixed
for contract execution it will varying depending on specific inputs

vikmeup: can we do `gasLimit * gasPrice` in this case? we should avoid hardcoding any values

dustinxie: fixed in latest push

PR #114, opened 2019-05-15
Add support for Ontology transactions Fixes #94

### LINKED ISSUE #94: Add support for Ontology transactions
`/transactions/AY6pRxUZKopz3rdxwy4yReeZgfBGA2ue48` - regular ONT transfers
`/transaction/AY6pRxUZKopz3rdxwy4yReeZgfBGA2ue48?token=ong` - should return token transfers for ong

ArtemGoryunov:
Please follow this checklist:

 - `platform/mycoin`
   - [x] `model.go`: Platform-specific models
   - [x] `client.go`: Platform getter methods (API, RPC)
   - [x] `api.go`:
     - Gin route: _GET /mycoin/txs_
     - Getting blockchain info
     - Normalizing platform-specific
       models to BlockAtlas model
   - [x] `api_test.go`
     - Unit tests
     - Check if normalization from
       response body matches expected model
 - [x] `test/main.go`:
   Add example address for integration test
 - [x] `loaders.go`: Add route
 - [x] `config.yml`: Add default config
   (comment out if no public endpoint)
 - [x] `config.go`: Add default config

### DIFF
--- a/platform/ontology/api.go
+++ b/platform/ontology/api.go
@@ -1,0 +1,117 @@
+package ontology
+
+import (
+	"github.com/gin-gonic/gin"
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/viper"
+	"github.com/trustwallet/blockatlas/coin"
+	"github.com/trustwallet/blockatlas/models"
+	"github.com/trustwallet/blockatlas/util"
+	"net/http"
+	"strings"
+)
+
+var client = Client{
+	HTTPClient: http.DefaultClient,
+}
+
+const GovernanceContract = "AFmseVrdL9f9oyCzZefL9tG6UbviEH9ugK"
+
+// Setup registers the Ontology route
+func Setup(router gin.IRouter) {
+	router.Use(util.RequireConfig("ontology.api"))
+	router.Use(func(c *gin.Context) {
+		client.BaseURL = viper.GetString("ontology.api")
+		c.Next()
+	})
+	router.GET("/:address", getTransactions)
+}
+
+func getTransactions(c *gin.Context) {
+	var token = c.DefaultQuery("token", "ont")
+	var address = c.Param("address")
+
+	txPage, error := client.GetTxsOfAddress(address, token, 1)
+
+	if error != nil {
+		logrus.WithError(error).
+			Errorf("Ontology: Failed to get transactions for %s, token %s", address, token)
+	}
+
+	var txs []models.Tx
+	for _, tx := range txPage.Result.TxnList {
+		if txNormalized, ok := Normalize(&tx, token); ok == true {
+			txs = append(txs, txNormalized)
+		}
+	}
+
+	page := models.Response(txs)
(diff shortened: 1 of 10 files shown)


vikmeup:
Tests? for both ont and ong

platform/ontology/client.go
```diff
@@ -0,0 +1,38 @@
...
+		page,
+	)
+
+	res, err := c.HTTPClient.Get(uri)
+	if err != nil {
+		logrus.WithError(err).Errorf("Ontology: Failed to get transactions for address %s for asset %s", address, assetName)
+	}
+	defer res.Body.Close()
+
+	print(res)
```
vikmeup: can you remove `print(res)` ?

platform/ontology/api.go
```diff
@@ -0,0 +1,117 @@
...
+		ID: srcTx.TxnHash,
+		Coin: coin.ONT,
+		Fee: models.Amount(fee),
+		Date:  srcTx.TxnTime,
+		Block: srcTx.Height,
+		Status: status,
+	}
+
+	// Condition for transfer ONT
+	if assetName == "ont" {
```
vikmeup: where would you get `ont` value from? user will be requesting `/transactions/:address`

platform/ontology/api.go
```diff
@@ -0,0 +1,117 @@
...
+		i := strings.IndexRune(transfer.Amount, '.')
```
vikmeup: what does this do: `strings.IndexRune(transfer.Amount, '.')` ?

platform/ontology/api.go
ArtemGoryunov: it's here 
https://github.com/TrustWallet/blockatlas/blob/1cef2ded64f0d831ff34a433e4986c629d332b98/platform/ontology/api.go#L31

vikmeup: I see now, can you have it as constant variables? because it's reused

platform/ontology/api.go
ArtemGoryunov: API returns amount value for ONT/ONG as string like `2.000000000`
ONT is a whole number and `strings.IndexRune(transfer.Amount, '.')` takes whole part.

platform/ontology/client.go
ArtemGoryunov: done

platform/ontology/api.go
ArtemGoryunov: Done

platform/ontology/api.go
```diff
@@ -16,6 +16,8 @@ var client = Client{
 }
 
 const GovernanceContract = "AFmseVrdL9f9oyCzZefL9tG6UbviEH9ugK"
+const ONTAssetName = "ont"
+const ONGAssetName = "ong"
```
vikmeup: check this file, that's how you specify consts: https://golang.org/src/math/const.go

PR #122, opened 2019-05-29
IoTeX get latest 25 txs of address

dustinxie:


platform/iotex/model.go
```diff
@@ -1,5 +1,17 @@
 package iotex
 
+type AccountInfo struct {
+	AccountMeta *AccountMeta `json:"accountMeta"`
+}
+
+type AccountMeta struct {
```
vikmeup: what is the purpose for account info model? i don't think it's needed

dustinxie: it is to query the total number of txs for the address, to compute proper start value in api.go

for example, totalNumber = 34, then ?start=9&count=25
if totalNumber = 14, then ?start=0&count=25

dustinxie: see changes in getTransactions() in api.go

vikmeup: can you make iotex api just return transactions in specific order? in this case we don't need to do an extra request

dustinxie: unfortunately we cannot, API is already defined and used in prod. endpoint and front-end for long time

i understand this is an extra request, but it's a simple HTTP GET which took roughly 200ms from our website statistics, it should not be a big deal?

PR #135, opened 2019-06-22
Improved coin registry

terorie:
* `coins.yml` instead of `coins.json`
 * Less places need changes for a coin integration
 * Improved CI tests
   * Integration test that starts a BA instance and requests each coin


coins.yml
```diff
@@ -0,0 +1,212 @@
+- id: 60
+  index: 0x8000003c
+  symbol: ETH
+  trust: ethereum
+  name: Ether
+  link: https://ethereum.org/ether
+  decimals: 18
+  sample: '0xfc10cab6a50a1ab10c56983c80cc82afc6559cf1'
```
vikmeup: did you consider to rename to `sampleAddress` ? we might do more samples in the future

coins.yml
```diff
@@ -0,0 +1,212 @@
...
+- id: 118
+  index: 0x80000076
+  symbol: ATOM
+  trust: cosmos
+  name: Cosmos ATOM
+  link: https://cosmos.network/
+  decimals: 6
+  sample: 'cosmos1rw62phusuv9vzraezr55k0vsqssvz6ed52zyrl'
+
+- id: 144
```
vikmeup: did you consider calling id as `slip44` and then rename `trust` to `id` ?

coins.yml
```diff
@@ -0,0 +1,212 @@
...
+  index: 0x80000334
+  symbol: CLO
+  trust: callisto
+  name: Callisto
+  link: http://callisto.network/
+  decimals: 18
+  sample: '0x39ec1c88a7a7c1a575e8c8f42eff7630d9278179'
+
+- id: 889
+  index: 0x80000379
```
vikmeup: what is the index?

terorie: Not sure, I got it from https://github.com/satoshilabs/slips/blob/master/slip-0044.md column `hexa`. Let's remove it

coins.yml
terorie: `id` isn't a good name imo, because it gets referenced as `handler` in the rest of the code. So how about either `trust`, `trustwallet` or `handler`.

vikmeup:
@kolya182 sounds like we need to rebase and apply changes: 

1) change `id` to `slip44` and then rename `trust` to `id` ?
2) remove `index` and `link`
3) rename `address` to `sampleAddress`

coins.yml
terorie: Renamed to `handler`

terorie:
Changes:
- Included changes from #143 
- @vikmeup concerns addressed
- `iotex` & few other bugfixes
- Code cleanup (especially `vechain`)
- New generic `TokenTxAPI` platform interface

coins.yml
terorie: Renamed to `handler`

terorie:
PR is ready to merge 🚀 Unit & Integration tests pass

PR #137, opened 2019-06-24
Waves integration

Tolsi:


kolya182:
@Tolsi Thank you for PR. I made a few changes, and remove token logic, once client supports it we will add it back + need to make additional changes to make it work properly.

PR #150, opened 2019-07-08
Binance dex transactions

kolya182:


platform/binance/client.go
```diff
@@ -88,6 +89,26 @@ func (c *Client) GetTxsOfAddress(address string, token string) (*TxPage, error)
 	return stx, nil
 }
 
+func (c *Client) GetTransactionReceipt(hash string) (*Receipt, error) {
+	url := fmt.Sprintf("%s/tx/%s?format=json", c.RPCBaseURL, hash)
+	println(url)
```
vikmeup: remove print

platform/binance/client.go
```diff
@@ -88,6 +89,26 @@ func (c *Client) GetTxsOfAddress(address string, token string) (*TxPage, error)
...
+	res, err := c.HTTPClient.Get(url)
+	if err != nil {
+		return nil, err
+	}
+	defer res.Body.Close()
+
+	recp := new(Receipt)
+	err = json.NewDecoder(res.Body).Decode(recp)
+	if err != nil {
```
vikmeup: I think you can just return err here and recp, you don't need to have another handle error

platform/binance/api.go
```diff
@@ -102,17 +108,94 @@ func NormalizeTx(srcTx *Tx, token string) (tx blockatlas.Tx, ok bool) {
 		return tx, true
 	}
 
+	// Conditin where explorer does not return sender/recepient for multisend transaction
```
vikmeup: `Conditin` => `Condition`

platform/binance/api.go
```diff
@@ -102,17 +108,94 @@ func NormalizeTx(srcTx *Tx, token string) (tx blockatlas.Tx, ok bool) {
...
+	if (srcTx.FromAddr == "" || srcTx.ToAddr == "") && srcTx.Type == TRANSFER {
+		println("hash :", hash)
```
vikmeup: remove println

platform/binance/api.go
```diff
@@ -102,17 +108,94 @@ func NormalizeTx(srcTx *Tx, token string) (tx blockatlas.Tx, ok bool) {
...
+		receipt, _ := p.client.GetTransactionReceipt(hash)
+		zeroMsgValue := receipt.TxReceipts.Value.Msg[0].MsgValue
+		zeroInput := zeroMsgValue.Inputs[0]
+		outputs := zeroMsgValue.Outputs
+		zeroOutputAdress := outputs[0].Address
```
vikmeup: do you have a test for it?

platform/binance/api.go
```diff
@@ -102,17 +108,94 @@ func NormalizeTx(srcTx *Tx, token string) (tx blockatlas.Tx, ok bool) {
...
+			tx.To = address
+			tx.From = zeroOutputAdress
+			tx.Meta = blockatlas.Transfer{
+				Value: blockatlas.Amount(coin.Amount),
+			}
+			return tx, true
+		}
+
+		// Condition for token_transfer in multisend transaction
+		if zeroInput.Coins[0].Denom != "BNB" {
```
vikmeup: why this condition? could be something else?

Issue #151, opened 2019-07-08
Add support for getting getting list of validators


vikmeup:
GET `/cosmos/validators`

```
[
   "docs" : [
      {
         "name":"Polychain Labs",
         "description":"Secure staking with Polychain Labs, the most experienced institutional grade staking team.",
         "status":"online/offline",
         "uptime": 100,
         "rate": 0.2,
         "info":{
            "website":"https://google.com",
            "image":"https://google.com/placeholder.png"
         },
         "address":"cosmosvaloper14k4pzckkre6uxxyd2lnhnpp8sngys9m6hl6ml7",
         "pubkey":"cosmosvalconspub1zcjduepquhlqdhjw4qp2c2t6qh5z7tfk52qc72623f0etc8f3n7hy8uuh25ql34fvu"
      }
   ],
   "status" : true
]
```

Example: https://api.cosmostation.io/v1/staking/validators

PR #152, opened 2019-07-08
Add staking validators

vikmeup:
Fixes: https://github.com/TrustWallet/blockatlas/issues/151

TODO:
- [x] Complete transaction format according to https://github.com/TrustWallet/blockatlas/issues/151
- [x] Forward `v1/:coin/:address ` to `v1/:coin/transactions/:address`
- [x] Clean up code for review

### DIFF
--- a/cmd/api/generic.go
+++ b/cmd/api/generic.go
@@ -6,7 +6,15 @@ import (
 	"net/http"
 )
 
-func makeTxRoute(router gin.IRouter, api blockatlas.Platform) {
+func makeTxRouteV1(router gin.IRouter, api blockatlas.Platform) {
+	makeTxRoute(router, api, "/:address")
+}
+
+func makeTxRouteV2(router gin.IRouter, api blockatlas.Platform) {
+	makeTxRoute(router, api, "/transactions/:address")
+}
+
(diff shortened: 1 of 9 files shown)


cmd/api/generic.go
```diff
@@ -16,7 +16,7 @@ func makeTxRoute(router gin.IRouter, api blockatlas.Platform) {
 		return
 	}
 
-	router.GET("/:address", func(c *gin.Context) {
+	router.GET("/transactions/:address", func(c *gin.Context) {
```
terorie: @vikmeup Moving the transactions route is quite a breaking change. How about doing it on a `/v2` route instead?

staking.go
```diff
@@ -0,0 +1,24 @@
...
+}
+
+const ValidatorsPerPage = 100
+
+type StakeValidatorInfo struct {
+	Name     string `json:"website"`
+	Description      string `json:"image"`
+}
+
+type StakeValidator struct {
```
terorie: Missing `go fmt`

cmd/api/generic.go
vikmeup: `v2` sounds good!

PR #168, opened 2019-07-16
Tuneup collectibles

kolya182:
- [x] Set renamed `collections_api` => `opensea_api` on Heroku . Ready to merge.


vikmeup:
I would prefer to use collectibles_api for consistency

PR #187, opened 2019-07-23
Support AE transactions 

eddywm:


platform/aeternity/api_test.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	dstJSON, err := json.Marshal(_test.expected)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if !bytes.Equal(resJSON, dstJSON) {
+		println(string(resJSON))
+		println(string(dstJSON))
```
vikmeup: import this lib, it's already part of the code:
`"github.com/stretchr/testify/assert"`
and then `assert.Equal(t, resJSON, dstJSON)`

go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/trustwallet/blockatlas
 go 1.12.6
 
 require (
+	github.com/davecgh/go-spew v1.1.1
```
vikmeup: an extra?

eddywm: It added itself when I called the assert.Equal dep, thought it was an indirect dependency.

I'm removing it.

platform/aeternity/api.go
```diff
@@ -0,0 +1,53 @@
...
+func NormalizeTx(srcTx *Transaction) blockatlas.Tx {
+	return blockatlas.Tx{
+		ID:     srcTx.Hash,
+		Coin:   coin.AE,
+		From:   srcTx.TxValue.Sender,
+		To:     srcTx.TxValue.Recipient,
+		Fee:    blockatlas.Amount(srcTx.TxValue.Fee),
+		Date:   int64(srcTx.Timestamp) / 1000,
+		Block:  srcTx.BlockHeight,
+		Memo:   "",
```
hewigovens: We can try base64 decoding `srcTx.payload` and if it's readable string, we can set it as `Memo`. e.g.: remove `ba_` and last 4 bytes checksum from `ba_SGVsbG8sIE1pbmVyISAvWW91cnMgQmVlcG9vbC4vKXcQag==` and decode, you will get `Hello, Miner! /Yours Beepool./`

platform/aeternity/client.go
```diff
@@ -0,0 +1,50 @@
...
+	if decodeError != nil {
+		return nil, decodeError
+	}
+	if len(txsArrays) == 0 {
+		return make([]Transaction, 0), nil
+	}
+
+	var result []Transaction
+	for _, tx := range txsArrays {
+		if tx.TxValue.Type == "SpendTx" {
```
hewigovens: Can we have enum in go?

vikmeup: haha, I wish had enum, I cry every time I want to use them

hewigovens: how about this https://blog.learngoprogramming.com/golang-const-type-enums-iota-bc4befd096d3, looks like no built-in support

platform/aeternity/api.go
eddywm: Can you link me to the docs, seems like the checksum might be bigger than 4 bytes
@hewigovens

eddywm: > Can you link me to the docs, seems like the checksum might be bigger than 4 bytes
> @hewigovens
nvm, I actually find out it's 8 bytes checksum.
https://github.com/aeternity/aepp-sdk-python/pull/204/files

hewigovens: I didn't find the docs, here is code shows how to encode the `payload`:

- https://github.com/aeternity/aepp-sdk-js/blob/73552e5aee/es/tx/builder/index.js#L54
- https://github.com/aeternity/aepp-sdk-js/blob/73552e5aee/es/utils/crypto.js#L161

platform/aeternity/client.go
eddywm: > how about this https://blog.learngoprogramming.com/golang-const-type-enums-iota-bc4befd096d3, looks like no built-in support

I think it might not be worth it, it looks hack-ish and here we just need to support one type

platform/aeternity/api.go
```diff
@@ -0,0 +1,73 @@
...
+	p.client.HTTPClient = http.DefaultClient
+	return nil
+}
+
+func (p *Platform) Coin() coin.Coin {
+	return coin.Coins[coin.AE]
+}
+
+func (p *Platform) GetTxsByAddress(address string) (blockatlas.TxPage, error) {
+	addressTxs, err := p.client.GetTxs(address, 25)
```
kolya182: @eddywm Would be more sense not to pass limit value but and reuse default limit from https://github.com/TrustWallet/blockatlas/blob/master/tx.go#L20 when construct url. Currently we don't pass `limit` param for transaction fetching.

platform/aeternity/api.go
```diff
@@ -0,0 +1,73 @@
...
+		txs = append(txs, NormalizeTx(&srcTx))
+	}
+	return txs, nil
+}
+
+func NormalizeTx(srcTx *Transaction) blockatlas.Tx {
+	return blockatlas.Tx{
+		ID:     srcTx.Hash,
+		Coin:   coin.AE,
+		From:   srcTx.TxValue.Sender,
```
kolya182: srcTx.TxValue can be reused

PR #193, opened 2019-07-26
Change Binance DEX config url

kolya182:
Load balancer points to `.../api/`  and in order to utilize private env var `ATLAS_BINANCE_DEX` current changes required.


vikmeup:
Current implementation is correct, not changes required for this.

PR #202, opened 2019-07-28
Fix duplicates when send to self and Token Transfers.

madcake:


test/tx_test.go
```diff
@@ -0,0 +1,61 @@
...
+)
+
+var transferDst = blockatlas.Tx{
+	ID:     "1681EE543FB4B5A628EF21D746E031F018E226D127044A4F9BA5EE2542A44555",
+	Coin:   coin.BNB,
+	From:   "tbnb1fhr04azuhcj0dulm7ka40y0cqjlafwae9k9gk2",
+	To:     "tbnb1sylyjw032eajr9cyllp26n04300qzzre38qyv5",
+	Fee:    "125000",
+	Date:   1555049867,
+	Block:  7761368,
```
vikmeup: is there a way to re-use this mocked transactions? it's duplicated in a few places? maybe create test package for this?

tx.go
```diff
@@ -146,3 +148,38 @@ func (t *Tx) GetAddresses() (addresses []string) {
...
 	}
 }
+
+type TxSet struct {
+	items map[Tx]bool
+	lock  sync.RWMutex
+}
+
+// Add adds a new element to the Set. Returns a pointer to the Set.
+func (s *TxSet) Add(t Tx) *TxSet {
```
vikmeup: would be better to have a separate file for TxSet

PR #221, opened 2019-07-31
Implement zilliqa observer

### LINKED ISSUE #208: [Observer] Add Zilliqa support for BlockAPI
Docs: https://apidocs.zilliqa.com/#getrecenttransactions

hewigovens:
Fixes #208, changes: 
- Add another json rpc client
- Get tx block height
-  Send batch json rpc requests to get txs
- Encode public key / key hash to bech32 zil address


cmd/config.go
```diff
@@ -68,4 +68,6 @@ func loadDefaults() {
 	viper.SetDefault("iotex.api", "https://pharos.iotex.io/v1")
 	viper.SetDefault("waves.api", "https://nodes.wavesnodes.com")
 	viper.SetDefault("aeternity.api", "https://mdw.aepps.com/")
+	viper.SetDefault("zilliqa.api", "https://api.viewblock.io/v1/zilliqa")
```
vikmeup: + `zilliqa.api_key` ?

go.mod
```diff
@@ -3,8 +3,14 @@ module github.com/trustwallet/blockatlas
 go 1.12.6
 
 require (
+	github.com/btcsuite/btcd v0.0.0-20190629003639-c26ffa870fd8 // indirect
+	github.com/btcsuite/btcutil v0.0.0-20190425235716-9e5f4b9a998d
```
vikmeup: I think you only need to include `github.com/btcsuite/btcutil` here ?

platform/zilliqa/model.go
```diff
@@ -12,3 +21,70 @@ type Tx struct {
...
+
+type TxRPC struct {
+	ID       string `json:"ID"`
+	Amount   string `json:"amount"`
+	GasLimit string `json:"gasLimit"`
+	GasPrice string `json:"gasPrice"`
+	Nonce    string `json:"nonce"`
+	Receipt  struct {
+		CumulativeGas string `json:"cumulative_gas"`
+		EpochNum      string `json:"epoch_num"`
```
vikmeup: can you create a separate struct for `Receipt` ?

platform/zilliqa/model.go
```diff
@@ -12,3 +21,70 @@ type Tx struct {
...
+		Success       bool   `json:"success"`
+	} `json:"receipt"`
+	SenderPubKey string `json:"senderPubKey"`
+	Signature    string `json:"signature"`
+	ToAddr       string `json:"toAddr"`
+	Version      string `json:"version"`
+}
+
+func (t *TxRPC)toTx() Tx {
```
vikmeup: can you implement this in the api.go as `NormalizeTx` ?

platform/zilliqa/model.go
```diff
@@ -12,3 +21,70 @@ type Tx struct {
...
+		Fee: fee.String(),
+		Signature: t.Signature,
+		Nonce: nonce,
+		ReceiptSuccess: t.Receipt.Success,
+	}
+	return tx
+}
+
+func encodePublicKeyToAddress(hexString string) string {
+	if strings.HasPrefix(hexString,"0x") {
```
vikmeup: can you add tests for this?

platform/zilliqa/model.go
```diff
@@ -12,3 +21,70 @@ type Tx struct {
...
+	}
+	bytes, err := hex.DecodeString(hexString)
+	if err != nil {
+		return ""
+	}
+	keyHash := sha256.Sum256(bytes)
+	return encodeKeyHashToAddress(keyHash[12:])
+}
+
+func encodeKeyHashToAddress(keyHash []byte) string {
```
vikmeup: I would suggest having a crypto package to have this method included there?

platform/zilliqa/model.go
```diff
@@ -12,3 +21,70 @@ type Tx struct {
...
+	conv, err := bech32.ConvertBits(keyHash, 8, 5, true)
+	if err != nil {
+		return ""
+	}
+	encoded, err := bech32.Encode("zil", conv)
```
vikmeup: can you add `zil` as a const?

platform/zilliqa/model.go
vikmeup: + add tests for normalizing func

vikmeup:
@hewigovens congrats with a first PR in go lang!

cmd/config.go
hewigovens: api key is set in env variable

platform/zilliqa/model.go
hewigovens: This `Receipt` is used only here, I would rather keep it nested

platform/zilliqa/model.go
hewigovens: We have 3 models, `blockatlas.Tx`, `zilliqa.Tx` and `zilliqa.TxRPC`, I want to separate the concern, in `api.go` we only deal with first two.

I can add a test for converting `TxRPC` to `Tx`

PR #238, opened 2019-08-03
Return IoTex transactions in a block

### LINKED ISSUE #213: [Observer] Add IoTeX support for BlockAPI
`v1/transfers/block/222669`

hewigovens:
Fixes #213

### DIFF
--- a/platform/iotex/api.go
+++ b/platform/iotex/api.go
@@ -7,7 +7,7 @@ import (
 	"time"
 
 	"github.com/spf13/viper"
-
+	"fmt"
 	"github.com/trustwallet/blockatlas/coin"
 )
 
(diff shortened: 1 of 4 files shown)


platform/iotex/api.go
```diff
@@ -25,10 +24,37 @@ func (p *Platform) Coin() coin.Coin {
...
+
+func (p *Platform) GetBlockByNumber(num int64) (*blockatlas.Block, error) {
+	var normalized []blockatlas.Tx
+	txs, err := p.client.GetTxsInBlock(num)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, action := range txs {
+		tx := Normalize(action)
```
vikmeup: Can you use go funk lib to do filter? Might be easier

hewigovens: The funk lib doesn't look good, prefer just for loop

PR #247, opened 2019-08-09
Support Nebulas (NAS) transactions

### LINKED ISSUE #235: Add support for Nebulas

ArtemGoryunov:
Fixes #235


platform/nebulas/client.go
```diff
@@ -0,0 +1,44 @@
...
+
+type Client struct {
+	HTTPClient *http.Client
+	URL        string
+}
+
+func (c *Client) GetTxs(address string, page int) ([]Transaction, error) {
+
+	uri := fmt.Sprintf("%s/tx?a=%s&p=%d",
+		c.URL,
```
vikmeup: Can you check Binance or waves client? We have a special method to request data easily

platform/nebulas/client.go
```diff
@@ -0,0 +1,44 @@
...
+		address,
+		page)
+	res, err := http.Get(uri)
+	if err != nil {
+		return nil, err
+	}
+
+	defer res.Body.Close()
+
+	if res.StatusCode != 200 {
```
kolya182:
```suggestion
	if res.StatusCode != http.StatusOK {
```

platform/nebulas/client.go
```diff
@@ -0,0 +1,44 @@
...
+		return nil, fmt.Errorf("http %s", res.Status)
```
kolya182:
```suggestion
		return nil, fmt.Errorf("Nebulas : %s", res.Status)
```

platform/nebulas/client.go
```diff
@@ -0,0 +1,44 @@
...
+	}
+
+	var response Response
+	if decodeError := json.NewDecoder(res.Body).Decode(&response); decodeError != nil {
+		return nil, decodeError
+	}
+
+	var result []Transaction
+	for _, tx := range response.Data.TxnList {
+		if tx.Type == "binary" {
```
kolya182: Can you create const of type operations instead hardcode ?

ArtemGoryunov: Done

PR #267, opened 2019-08-17
Add bitcoin and forks block time

### LINKED ISSUE #266: [Observer] Add proper block time for each coin in coins.yml

hewigovens:
Fixes #266, and the min poll is also changed to 1250 ms = 5s / 4

### DIFF
--- a/coins.yml
+++ b/coins.yml
@@ -73,6 +73,7 @@
   handle: iotex
   name: IoTeX
   decimals: 18
+  blockTime: 10000
   sampleAddress: 'io1mwekae7qqwlr23220k5n9z3fmjxz72tuchra3m'
 
 - id: 313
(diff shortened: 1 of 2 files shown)


coins.yml
```diff
@@ -210,81 +213,93 @@
   handle: bitcoin
   name: Bitcoin
   decimals: 8
+  blockTime: 600000
```
vikmeup: can we do `10000` - this means it will check every 5-10 seconds for the new transactions

hewigovens: No new blocks, seems we need to check mempool

vikmeup: I don't think we need an infra in place for it, currently it just checks confirmed blocks, we can add this logic in the future version

PR #271, opened 2019-08-22
[Observer] Add support for bitcoin like blockchains to observe transactions

vikmeup:


hewigovens:
Let's register xpub instead, to avoid sending multiple notifications (e.g.: a tx has multiple inputs)

Issue #249, opened 2019-08-09
[Observer] Add Nebulas (NAS) support for BlockAPI


ArtemGoryunov:
Documentation https://github.com/nebulasio/wiki/blob/master/rpc.md
There are the following methods:

```
LatestIrreversibleBlock
GetBlockByHash
GetBlockByHeight
```

Lekky71:
Hi @ArtemGoryunov 
I have a few questions concerning the Nebula client.
1. There is an existing `GetTxs` function that calls the 'tx' endpoint but this endpoint is not in the documentation shared via the link above, is it not valid or there is another documentation?

2. When the Block response is gotten from the http request, which data is needed to be filtered out, 
Is it the entire response, or just the transactions list contained in it?

I look forward to your response, thank you.

vikmeup:
@Lekky71 take a look at https://github.com/trustwallet/blockatlas/issues/268, I will be updating information on blockatlas usage, there is good examples.

Lekky71:
Alright, thanks, I looked at the example PRs and saw that I have already done the same with the nebula platform, the only difference is that the methods are different (LatestIrreversibleBlock, GetBlockByHash, GetBlockByHeight), how do I hook them up with the observer?

PR #272, opened 2019-08-23
[Observer] Add Nebulas (NAS) support for BlockAPI #249

akinmail:
The existing GetTxsByAddress api seems to be using a different api from the one in the documentation you sent us. There are subtle differences in the models you already had(Transaction) and the one in the documentation, for example "nonce" is not included in a transaction object, rather its in the block object.

Another example is "Response" from the api has "result" instead of data you were using as the json key for GetTxsByAddress.

This is why i had to create my own models that just extended yours. Hope that's fine?

I am available to explain other decisions i made in the code.

Fixes: https://github.com/trustwallet/blockatlas/issues/249


platform/nebulas/api.go
```diff
@@ -56,3 +77,41 @@ func NormalizeTx(srcTx Transaction) blockatlas.Tx {
 		},
 	}
 }
+
+func NormalizeNasTx(srcTx NasTransaction, block NasBlock) blockatlas.Tx {
```
ArtemGoryunov: This is the simular to https://github.com/trustwallet/blockatlas/blob/30f6ace4f32d7d5212b2ea9ad13c4a268aaa6802/platform/nebulas/api.go#L58

platform/nebulas/model.go
```diff
@@ -23,6 +33,13 @@ type Transaction struct {
 	Status      int32       `json:"status"`
 }
 
+type NasTransaction struct {
```
ArtemGoryunov: Can be `NasTransaction` and `Transaction` merged into one?

platform/nebulas/api.go
akinmail: An existing GetTxsByAddress() api in this file uses the function NormalizeTx() and it seems to be calling a different api endpoint not included in the documentation you sent us. here https://github.com/nebulasio/wiki/blob/master/rpc.md

So since I don't have the documentation of the api that GetTxsByAddress() api is calling, I didn't want to break the code that's why I created another that normalize specifically for my api.

But I will remove the new one i created and just update the old one that's there.

platform/nebulas/model.go
akinmail: Yes will do that

platform/nebulas/api.go
akinmail: Done

platform/nebulas/model.go
akinmail: Done

akinmail: Done

platform/nebulas/api.go
akinmail: I will include a test for this new normalizer function using its own sample inputs and outputs and you will see that the format is different from the first one.

platform/nebulas/model.go
akinmail: No could not merge them as the data type of some of the fields were different, address is a struct in the first case but a string in the second case

platform/nebulas/api.go
akinmail: Added test for new normalizer function

config.yml
```diff
@@ -135,7 +135,8 @@ aeternity:
 
 # [NAS] Nebulas: https://nebulas.io
 nebulas:
-  api: https://mainnet.nebulas.io
+  api: https://explorer-backend.nebulas.io/api
+  mainnet: https://mainnet.nebulas.io
```
vikmeup: can you call it `rpc`, instead of mainnet?

PR #275, opened 2019-08-25
Add bitcoin block api

hewigovens:


platform/bitcoin/client.go
```diff
@@ -26,22 +27,46 @@ func InitClient(URL string) Client {
...
+	err = c.Request.Get(&block, c.URL, path, args)
+	return block, err
+}
+
+func (c *Client) GetTransactionsByBlockChan(number int64, page int64, out chan Block, wg *sync.WaitGroup) {
+	wg.Add(1)
+	defer wg.Done()
+	block, err := c.GetTransactionsByBlock(number, page)
+	if err != nil {
+		return
```
Pantani: We need handler this error. My suggestion is to create another channel to handle errors. Or we can put the result inside a struct with the `block` object and the error.

platform/bitcoin/api_test.go
```diff
@@ -114,8 +124,18 @@ func TestInferDirection(t *testing.T) {
 	set.Add("bc1q2sykr9c342mjpm9mwnps8ksk6e35lz75rpdlfe")
 	set.Add("bc1qs86ucvr3unce2grvfp77433npy66nzha9w0e3c")
 
-	inputs1 := []string{"bc1q2sykr9c342mjpm9mwnps8ksk6e35lz75rpdlfe"}
-	outputs1 := []string{"bc1q6wf7tj62f0uwr6almah3666th2ejefdg72ek6t"}
+	inputs1 := []blockatlas.TxOutput{
+		blockatlas.TxOutput{
```
Pantani: We don't need to define the type again.

```suggestion
		{
```

platform/bitcoin/api_test.go
```diff
@@ -114,8 +124,18 @@ func TestInferDirection(t *testing.T) {
...
+			Address: "bc1q2sykr9c342mjpm9mwnps8ksk6e35lz75rpdlfe",
+			Value:   "",
+		},
+	}
+	outputs1 := []blockatlas.TxOutput{
+		blockatlas.TxOutput{
```
Pantani: We don't need to define the type again.

```suggestion
		{
```

platform/bitcoin/api_test.go
```diff
@@ -126,16 +146,41 @@ func TestInferDirection(t *testing.T) {
 		t.Error("direction is not outgoing")
 	}
 
-	inputs2 := []string{
-		"3CgvDkzcJ7yMZe75jNBem6Bj6nkMAWwMEf",
-		"3LyzYcB54pm9EAMmzXpFfb1kzEDAFvqBgT",
-		"3Q6DYour5q5WdMhyXsyPgBeAqPCXchzCsF",
-		"3JZZM1rwst7G5izxbFL7KNvy7ZiZ47SVqG",
+	inputs2 := []blockatlas.TxOutput{
+		blockatlas.TxOutput{
```
Pantani: The same applies for all object in this array. We don't need to define the type again.

```suggestion
		{
```

platform/bitcoin/api_test.go
```diff
@@ -126,16 +146,41 @@ func TestInferDirection(t *testing.T) {
...
+			Address: "3JZZM1rwst7G5izxbFL7KNvy7ZiZ47SVqG",
+			Value:   "",
+		},
 	}
-	outputs2 := []string{
-		"139f1CrnLWvVajGzs3ZtpQhbGWxM599sho",
-		"3LyzYcB54pm9EAMmzXpFfb1kzEDAFvqBgT",
-		"bc1q9mx5tm66zs7epa4skvyuf2vfuwmtnlttj74cnl",
+	outputs2 := []blockatlas.TxOutput{
+		blockatlas.TxOutput{
```
Pantani: The same applies for all object in this array. We don't need to define the type again.

```suggestion
		{
```

platform/bitcoin/api_test.go
```diff
@@ -126,16 +146,41 @@ func TestInferDirection(t *testing.T) {
 		t.Error("direction is not outgoing")
 	}
 
-	inputs2 := []string{
-		"3CgvDkzcJ7yMZe75jNBem6Bj6nkMAWwMEf",
-		"3LyzYcB54pm9EAMmzXpFfb1kzEDAFvqBgT",
-		"3Q6DYour5q5WdMhyXsyPgBeAqPCXchzCsF",
-		"3JZZM1rwst7G5izxbFL7KNvy7ZiZ47SVqG",
+	inputs2 := []blockatlas.TxOutput{
```
Pantani: We can do a table test here, its easier and we don't need to make the logic again for more than one case.

This is a sample:
https://github.com/golang/go/wiki/TableDrivenTests

platform/bitcoin/api.go
```diff
@@ -89,6 +91,40 @@ func (p *Platform) getTxsByAddress(address string) ([]blockatlas.Tx, error) {
...
+		var wg sync.WaitGroup
+		out := make(chan Block)
+		for i := int64(2); i <= block.TotalPages; i++ {
+			go p.client.GetTransactionsByBlockChan(num, i, out, &wg)
+		}
+
+		wg.Wait()
+		defer close(out)
+		for r := range out {
+			block.Transactions = append(block.Transactions, r.Transactions...)
```
Pantani: Can we normalize the all transactions here? Because we don't need to iterate the array again to normalize the transactions.

platform/bitcoin/api.go
```diff
@@ -99,76 +135,96 @@ func NormalizeTxs(sourceTxs TransactionsList, coinIndex uint, addressSet mapset.
...
 	}
-	return result
+	for _, val := range ordered {
+		addresses = append(addresses, set[val])
+	}
+	return addresses
+}
+
+func addAmount(left string, right string) (sum blockatlas.Amount) {
+	amount1, _ := strconv.ParseInt(left, 10, 64)
```
Pantani: We need to handle this error, or just log them.

platform/bitcoin/api.go
```diff
@@ -99,76 +135,96 @@ func NormalizeTxs(sourceTxs TransactionsList, coinIndex uint, addressSet mapset.
...
+	amount2, _ := strconv.ParseInt(right, 10, 64)
```
Pantani: We need to handle this error, or just log them.

platform/bitcoin/api.go
hewigovens: we have tried, that will make code more tedious (one more loop)

PR #281, opened 2019-08-30
Add endpoint to list all available coins

Pantani:
## Summary

Add endpoint to list all available coins.

## Problem Definition

The new endpoint was going help to create the integration tests and other services.

## Proposal

Create new routes `/coins` to list all coins and `/coin/:id` to get a unique coin.

Response sample:
```
[
  {
    "ID": 0,
    "Handle": "bitcoin",
    "Symbol": "BTC",
    "Title": "Bitcoin",
    "Decimals": 8,
    "BlockTime": 600000,
    "SampleAddr": "bc1quvuarfksewfeuevuc6tn0kfyptgjvwsvrprk9d"
  }
]
```

### DIFF
--- a/cmd/api/generic.go
+++ b/cmd/api/generic.go
@@ -151,7 +151,11 @@ func makeTokenRoute(router gin.IRouter, api blockatlas.Platform) {
 
 func makeCoinRoute(router gin.IRouter) {
 	router.GET("/coins", func(c *gin.Context) {
-		c.JSON(http.StatusOK, coin.Coins)
+		coins := make([]coin.Coin, 0)
+		for _, coin := range coin.Coins {
+			coins = append(coins, coin)
+		}
+		c.JSON(http.StatusOK, coins)
 	})
 }
 
--- a/cmd/api/generic.go
+++ b/cmd/api/generic.go
@@ -7,7 +7,6 @@ import (
 	services "github.com/trustwallet/blockatlas/services/assets"
 	"log"
 	"net/http"
-	"strconv"
 )
 
 func makeTxRouteV1(router gin.IRouter, api blockatlas.Platform) {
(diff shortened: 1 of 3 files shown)


cmd/api/routes.go
```diff
@@ -41,6 +41,10 @@ func loadPlatforms(root gin.IRouter) {
 		customAPI.RegisterRoutes(router)
 	}
 
+	makeCoinRoute(root)
+	makeCoinRoute(getRouter(v1, ""))
```
hewigovens: I think we only need to register one route? v1 is enough, if we need to change something then we add v2

hewigovens:
It would be nice to create a review app (pr) in heroku, so we can test it

cmd/api/routes.go
vikmeup: root is enough I think

cmd/api/generic.go
```diff
@@ -148,6 +150,26 @@ func makeTokenRoute(router gin.IRouter, api blockatlas.Platform) {
 	})
 }
 
+func makeCoinRoute(router gin.IRouter) {
+	router.GET("/coins", func(c *gin.Context) {
+		c.JSON(http.StatusOK, coin.Coins)
+	})
+
+	router.GET("/coin/:id", func(c *gin.Context) {
```
vikmeup: what is the purpose for `/coin/:id` ? I think `router.GET("/coins"` might be enough

vikmeup: let's only add method that we will need

kolya182:
> It would be nice to create a review app (pr) in heroku, so we can test it

kolya182:
> It would be nice to create a review app (pr) in heroku, so we can test it

https://blockatlas-trust-staging.herokuapp.com/coins

cmd/api/generic.go
Pantani: I going to remove, I did use in the integrations test, but it is no more useful

hewigovens:
```json
{
    "0": {
        "ID": 0,
        "Handle": "bitcoin",
        "Symbol": "BTC",
        "Title": "Bitcoin",
        "Decimals": 8,
        "BlockTime": 600000,
        "SampleAddr": "bc1quvuarfksewfeuevuc6tn0kfyptgjvwsvrprk9d"
    }
}
```

I think we should lowercase the json field, `id`, `handle`... and maybe an array is more intuitive?

cmd/api/generic.go
```diff
@@ -148,6 +149,16 @@ func makeTokenRoute(router gin.IRouter, api blockatlas.Platform) {
 	})
 }
 
+func makeCoinRoute(router gin.IRouter) {
+	router.GET("/coins", func(c *gin.Context) {
+		coins := make([]coin.Coin, 0)
+		for _, coin := range coin.Coins {
+			coins = append(coins, coin)
+		}
+		c.JSON(http.StatusOK, coins)
```
vikmeup: is it possible just to return `c.JSON(http.StatusOK, []coin.Coin)` ?

Pantani: We can save the objects inside a slice in the coin package, to avoid the iteration every time we call this API. The structure inside a `map[uint]Coin` it is harder to parse

PR #291, opened 2019-09-02
Add type Direction and Status

vikmeup:


vikmeup:
Created separate issue to add for transaction type: https://github.com/trustwallet/blockatlas/issues/292

PR #301, opened 2019-09-03
Fix xpub webhook for tx receiver

Pantani:
# Headline
Fix webhook to send a request for transaction receiver for UTXO's coins.

## Problem
If you register a webhook for the sender and receiver, just the sender receives the webhook. Because to addresses to infer the direction are wrong.

## Solution
Save a list of addresses per xpub, to know the addresses, and infer the direction for tx

### DIFF
--- a/observer/observer.go
+++ b/observer/observer.go
@@ -46,12 +46,18 @@ func (o *Observer) processBlock(events chan<- Event, block *blockatlas.Block) {
 	emitted := make(map[string]blockatlas.Direction)
 	platform := bitcoin.UtxoPlatform(o.Coin)
 	for _, sub := range subs {
+
 		txs := txMap[sub.Address].Txs()
 		for _, tx := range txs {
+
 			xpub, _ := o.Storage.GetXpubFromAddress(o.Coin, sub.Address)
-			if len(xpub) != 0 {
+			if len(xpub) > 0 {
+				xpubAddresses, err := o.Storage.GetAddressFromXpub(o.Coin, xpub)
+				if err != nil {
+					continue
+				}
 				addressSet := mapset.NewSet()
-				for _, addr := range tx.GetUtxoAddresses() {
+				for _, addr := range xpubAddresses {
 					addressSet.Add(addr)
 				}
 				direction := platform.InferDirection(&tx, addressSet)
@@ -64,14 +70,16 @@ func (o *Observer) processBlock(events chan<- Event, block *blockatlas.Block) {
 					Decimals: coin.Coins[o.Coin].Decimals,
 				}
 			}
+
 			if d, ok := emitted[tx.ID]; ok {
-				if d == tx.Direction || d == blockatlas.DirectionSelf{
+				if d == tx.Direction || d == blockatlas.DirectionSelf {
 					continue
 				}
 				emitted[tx.ID] = blockatlas.DirectionSelf
 			} else {
 				emitted[tx.ID] = tx.Direction
 			}
+
 			events <- Event{
 				Subscription: sub,
 				Tx:           &tx,
--- a/observer/observer.go
+++ b/observer/observer.go
@@ -43,7 +43,7 @@ func (o *Observer) processBlock(events chan<- Event, block *blockatlas.Block) {
 	}
 
 	// Emit events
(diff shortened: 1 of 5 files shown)


observer/observer.go
```diff
@@ -67,7 +64,14 @@ func (o *Observer) processBlock(events chan<- Event, block *blockatlas.Block) {
 					Decimals: coin.Coins[o.Coin].Decimals,
 				}
 			}
-			emitted[tx.ID] = true
+			if d, ok := emitted[tx.ID]; ok {
+				if d == tx.Direction || d == blockatlas.DirectionSelf{
```
madcake: What’s different with just true?

Pantani: because tx going to send just one time per observer, now we filter by direction, the tx going to sent one time per direction and per observer

madcake: It’s for utxo only?

Pantani: yes

observer/storage/redis/storage.go
```diff
@@ -43,8 +44,34 @@ func (s *Storage) SaveXpubAddresses(coin uint, addresses []string, xpub string)
 	for _, address := range addresses {
 		a[address] = xpub
 	}
+
+	key := fmt.Sprintf(keyXpub, coin)
+	err := s.saveHashMap(key, a)
+	if err != nil {
+		return err
+	}
+	j, err := json.Marshal(addresses)
```
madcake: I think with this mongo more speedy

observer/observer.go
```diff
@@ -43,18 +43,21 @@ func (o *Observer) processBlock(events chan<- Event, block *blockatlas.Block) {
...
 	for _, sub := range subs {
+
 		txs := txMap[sub.Address].Txs()
 		for _, tx := range txs {
-			if _, ok := emitted[tx.ID]; ok {
-				continue
-			}
+
 			xpub, _ := o.Storage.GetXpubFromAddress(o.Coin, sub.Address)
-			if len(xpub) != 0 {
```
madcake: Could you out utxo handling

observer/observer.go
```diff
@@ -67,7 +70,16 @@ func (o *Observer) processBlock(events chan<- Event, block *blockatlas.Block) {
 					Decimals: coin.Coins[o.Coin].Decimals,
 				}
 			}
-			emitted[tx.ID] = true
+
+			if d, ok := emitted[tx.ID]; ok {
+				if d == tx.Direction || d == blockatlas.DirectionSelf {
```
madcake: More simple for read:
```go
if (d != tx.Direction && d != blockatlas.DirectionSelf) {
  emitted[tx.ID] = blockatlas.DirectionSelf
}
```

observer/observer.go
vikmeup: if for all, account based transactions does not have direction type, it's empty I assume it's just for utxo then?

Pantani: sorry, I thought all type of tx we infer the direction. I am going to change this

observer/observer.go
Pantani: Is the same verification, but we don't match the negative numbers

observer/observer.go
Pantani: but we need to call the `continue` to loop go to the next object and don't send the event again

observer/observer.go
```diff
@@ -43,18 +43,21 @@ func (o *Observer) processBlock(events chan<- Event, block *blockatlas.Block) {
...
 		for _, tx := range txs {
-			if _, ok := emitted[tx.ID]; ok {
-				continue
-			}
+
 			xpub, _ := o.Storage.GetXpubFromAddress(o.Coin, sub.Address)
-			if len(xpub) != 0 {
+			if len(xpub) > 0 {
+				xpubAddresses, err := o.Storage.GetAddressFromXpub(o.Coin, xpub)
+				if err != nil {
```
vikmeup: can we move out some of the logic and add unit testing? to make sure direction handled properly

Pantani: This tests already exist, the problem was the parameters passed to the method were wrong

https://github.com/trustwallet/blockatlas/blob/ad803284bceff444bc4dab9752dcc892e304e466/platform/bitcoin/api_test.go#L197

Issue #224, opened 2019-07-31
Dispatch: hook count = 0


vikmeup:
2019-07-31T23:28:10.164538+00:00 app[observer.1]: time="2019-07-31T23:28:10Z" level=info msg="Dispatch: hook count = 0" coin=144 txID=B60F539AACFCA19520FE9774692942F3BCCAF6F16047CC5FE47273D7B7923F46 webhook="[]"

Pantani:
#306

PR #306, opened 2019-09-04
Remove addresses without webhook from Redis

Pantani:
# Headline
Remove keys with empty webhooks from Redis to avoid useless processing

Issue [224](https://github.com/trustwallet/blockatlas/issues/224)

### DIFF
--- a/observer/storage/redis/storage.go
+++ b/observer/storage/redis/storage.go
@@ -146,6 +146,7 @@ func (s *Storage) Delete(subs []observer.Subscription) error {
 
 func (s *Storage) updateWebHooks(subs []observer.Subscription, operation webHookOperation) error {
 	fields := make(map[string]interface{})
+	del := make([]string, 0)
 	keys := make([]string, 0)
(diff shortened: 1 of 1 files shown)


Pantani:
> Should we call this method only if len(del) > 0?

Make sense! I am going to change

observer/storage/redis/storage.go
```diff
@@ -146,6 +146,7 @@ func (s *Storage) Delete(subs []observer.Subscription) error {
 
 func (s *Storage) updateWebHooks(subs []observer.Subscription, operation webHookOperation) error {
```
madcake: when is this called?

Pantani: [here](https://github.com/trustwallet/blockatlas/pull/306/files#diff-1696d28c4b8857cb3cda676c006721b4R140) and [here](https://github.com/trustwallet/blockatlas/pull/306/files#diff-1696d28c4b8857cb3cda676c006721b4R144)

Issue #234, opened 2019-08-02
Add retry mechanism when fetching txs in a block


hewigovens:
For example, zilliqa:

RPC might have some delay
```json
{
  "error": {
    "code": -32602,
    "data": null,
    "message": "INVALID_PARAMS: Invalid method parameters (invalid name and/or type) recognised: Tx Block does not exist"
  },
  "id": "1",
  "jsonrpc": "2.0"
}
```

after a few minutes

```json
{
  "error": {
    "code": -1,
    "data": null,
    "message": "TxBlock has no transactions"
  },
  "id": "1",
  "jsonrpc": "2.0"
}
```

Pantani:
#307

PR #307, opened 2019-09-05
Add retry for GetBlockByNumber method

Pantani:
# Headline
Add retry mechanism to the observer fetch transactions inside a block 

Issue [234](https://github.com/trustwallet/blockatlas/issues/234)

### DIFF
--- a/observer/retry.go
+++ b/observer/retry.go
@@ -1,0 +1,31 @@
+package observer
+
+import (
+	"github.com/trustwallet/blockatlas"
+	"math/rand"
+	"time"
+)
+
+type GetBlockByNumber func(num int64) (*blockatlas.Block, error)
+
+type stop struct {
+	error
+}
+
+func retry(attempts int, sleep time.Duration, f GetBlockByNumber, n int64) (*blockatlas.Block, error) {
+	r, err := f(n)
+	if err != nil {
+		if s, ok := err.(stop); ok {
+			return nil, s.error
+		}
+		if attempts--; attempts > 0 {
+			// Add some randomness to prevent creating a Thundering Herd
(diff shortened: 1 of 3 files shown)


observer/stream.go
```diff
@@ -87,7 +87,7 @@ func (s *Stream) loadBlock(c chan<- *blockatlas.Block, num int64) {
 	s.semaphore.Acquire()
 	defer s.semaphore.Release()
 
-	block, err := s.BlockAPI.GetBlockByNumber(num)
+	block, err := retry(3, time.Second*1, s.BlockAPI.GetBlockByNumber, num)
```
vikmeup: can we do 5 retries? it would do 1, then wait 5 sec, then 15 sec and then 30 sec and 60? any examples on how that works.

would be great to log errors on this, if some block failed

observer/retry.go
```diff
@@ -0,0 +1,31 @@
...
+		if s, ok := err.(stop); ok {
+			return nil, s.error
+		}
+		if attempts--; attempts > 0 {
+			// Add some randomness to prevent creating a Thundering Herd
+			jitter := time.Duration(rand.Int63n(int64(sleep)))
+			sleep = sleep + jitter/2
+
+			time.Sleep(sleep)
+			return retry(attempts, sleep*2, f, n)
```
madcake: Code formatting.

observer/stream.go
Pantani: done!

observer/retry.go
Pantani: This file is using the default format for Golang, `go fmt`. They automatically arrange the var with `*2` without spaces. This pattern is used by default in Golang.  If we prefer to follow this rule, I can change this for Goland stop to do this automatically. What do you think?

PR #325, opened 2019-09-08
filter ERC collections

Pantani:
# Headline
Filter all collections only for ERC721 and ERC1155

### DIFF
--- a/platform/ethereum/model.go
+++ b/platform/ethereum/model.go
@@ -1,5 +1,7 @@
 package ethereum
 
+import "math/big"
+
 type Page struct {
 	Total uint  `json:"total"`
 	Docs  []Doc `json:"docs"`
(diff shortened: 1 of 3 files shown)


platform/ethereum/model.go
```diff
@@ -61,7 +63,7 @@ type Collection struct {
 	ImageUrl    string                 `json:"image_url"`
 	Description string                 `json:"description"`
 	ExternalUrl string                 `json:"external_url"`
-	Total       int                    `json:"owned_asset_count"`
+	Total       *big.Int               `json:"owned_asset_count"`
```
vikmeup: Can we keep this as int?

Pantani: I got one case with the `owned_asset_count` equal to `20000000000000000000`, and we got an Unmarshal error

Issue #329, opened 2019-09-09
Collections without assets_contracts can cause panic


Pantani:
E.g.:
address: `0xE96a1B303A1eb8D04fb973eB2B291B8d591C8F72`

```
[
  {
    "primary_asset_contracts": [],
    "traits": {
      "Max_Bonus": {
        "min": 300,
        "max": 300
      },
      "Bonus_Per_Token": {
        "min": 1,
        "max": 1
      }
    },
    "stats": {
      "seven_day_volume": 0,
      "seven_day_change": -1,
      "total_volume": 0.04,
      "count": 3,
      "num_owners": 1,
      "market_cap": 0,
      "average_price": 0.0133333333333333,
      "items_sold": 3
    },
    "name": "Dissolution",
    "slug": "dissolution",
    "image_url": "https://storage.opensea.io/dissolution-1566503734.png",
    "short_description": null,
    "description": "Experience tactical FPS combat in a cutthroat universe ravaged by an ongoing war of extinction between humanity and AI. Fight for loot backed by blockchain in competitive game modes.",
    "external_url": "https://dissolution.online",
    "chat_url": null,
    "wiki_url": null,
    "large_image_url": "https://storage.opensea.io/dissolution-large-1566503735.png",
    "featured_image_url": null,
    "featured": false,
    "banner_image_url": null,
    "display_data": {
      "card_display_style": "contain",
      "images": [
        "https://storage.opensea.io/0xfaafdc07907ff5120a76b34b731b278c38d6043c/39803530675327460487158288219684256433559304725576879659134376004308812431360-1566015172.jpg"
      ]
    },
    "hidden": false,
    "created_date": "2019-08-22T19:31:10.369032",
    "owned_asset_count": 20
  },
  {
    "primary_asset_contracts": [],
    "traits": {},
    "stats": {
      "seven_day_volume": 0,
      "seven_day_change": 0,
      "total_volume": 0,
      "count": 189,
      "num_owners": 1,
      "market_cap": 0,
      "average_price": 0,
      "items_sold": 0
    },
    "name": "Spirit Clash",
    "slug": "spirit-clash",
    "image_url": "https://storage.opensea.io/spirit-clash-1566593935.png",
    "short_description": null,
[...]

PR #330, opened 2019-09-09
Filter collections without contracts

Pantani:
# Headline 
Filter and exclude collections without contract.

## Problem
Collections without assets_contracts can cause panic [#329](https://github.com/trustwallet/blockatlas/issues/329)


ArtemGoryunov:
@Pantani I asked some questions about `.Contracts.count` = 0. As I understand if we skip it some collectibles will be skipped too.

Issue #292, opened 2019-09-02
Improve transaction type


vikmeup:
We can create `type TransactionType string` to make it safer to use

https://github.com/trustwallet/blockatlas/blob/5768db89be55d07e2a87738750d62b45b6a17e17/tx.go#L79

Pantani:
#332

PR #332, opened 2019-09-10
create the Tx `TransactionType` 

Pantani:
Issue [292](https://github.com/trustwallet/blockatlas/issues/292)

### DIFF
--- a/observer/dispatcher.go
+++ b/observer/dispatcher.go
@@ -26,7 +26,7 @@ func (d *Dispatcher) Run(events <-chan Event) {
 
 func (d *Dispatcher) dispatch(event Event) {
 	action := DispatchEvent{
-		Action: event.Tx.Type,
+		Action: string(event.Tx.Type),
 		Result: event.Tx,
 	}
 	txJson, err := json.Marshal(action)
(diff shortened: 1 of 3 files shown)


tx.go
```diff
@@ -14,6 +14,7 @@ const (
 type Direction string
 type Status string
 type TokenType string
+type TransactionType string
```
vikmeup: would be good to add transactions types into constant?

observer/dispatcher.go
```diff
@@ -26,7 +26,7 @@ func (d *Dispatcher) Run(events <-chan Event) {
 
 func (d *Dispatcher) dispatch(event Event) {
 	action := DispatchEvent{
-		Action: event.Tx.Type,
+		Action: string(event.Tx.Type),
```
vikmeup: should action use `TransactionType ` ?

tx.go
Pantani: yes!! done!!

observer/dispatcher.go
Pantani: done!

PR #334, opened 2019-09-10
refector vechain api

Pantani:


Pantani:
> LGTM!
> 
> Do you think we should expose `/:chain/:block` endpoint? so we can do integration testing for it and might be useful for other use cases, because we already use `BlockAPI` to fetch every block fo observer

I think it is a good thing to put in the app, like a block explorer for all coins, and we can see some status about the last block. But if we won't use, Is just a more endpoint to maintain.

PR #335, opened 2019-09-10
Change all logs to new log package

Pantani:


kolya182:
github.com/sirupsen/logrus was replaced by logger, but i still see dependency in `go.mod`, do we really need it ?

Pantani:
> github.com/sirupsen/logrus was replaced by logger, but i still see dependency in `go.mod`, do we really need it ?

Yes, because we use the `Logrus` inside the logger package

PR #336, opened 2019-09-10
Fix integration tests

Pantani:
# Headline
- Change the address for IOTEX in the integration tests.
- Remove the collection's tests for coins `classic`, `callisto`, `poa`, `thundertoken`, `tomochain`, `gochain`.

## Problem
- The old IOTEX address has a rate limit issue because it has a lot of transactions.
- Sometimes the integration tests for collections API's fail because the Ethereum based coins don't have collections. If you try to use the same environment from Ethereum collections API, we have a rate limit as well.

## Solution
- Use an address with fewer transactions.
- Remove the coins `classic`, `callisto`, `poa`, `thundertoken`, `tomochain`, `gochain` for collection's tests.

### DIFF
--- a/pkg/integration/testdata/exclude.json
+++ b/pkg/integration/testdata/exclude.json
@@ -1,3 +1,14 @@
 [
-
+  "/v2/classic/collections/:owner",
+  "/v2/callisto/collections/:owner",
+  "/v2/poa/collections/:owner",
+  "/v2/thundertoken/collections/:owner",
+  "/v2/tomochain/collections/:owner",
+  "/v2/gochain/collections/:owner",
+  "/v2/classic/collections/:owner/collection/:collection_id",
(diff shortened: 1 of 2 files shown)


pkg/integration/testdata/exclude.json
```diff
@@ -1,3 +1,14 @@
 [
-
+  "/v2/classic/collections/:owner",
+  "/v2/callisto/collections/:owner",
+  "/v2/poa/collections/:owner",
```
vikmeup: maybe instead of excluding we add list of tests that needs to run? i'm ok either way

Pantani: I think to add the tests to exclude is more easy, because we don't need to map all routes, only exclude what doesn't make sense.

PR #339, opened 2019-09-11
Fix XRP block parse

Pantani:
# Headline

Some XRP transactions can be a transaction receipt about from other coins. Like the payload bellow, this is an ETH transaction inside the XRP ledger.

## Problem
We have a problem when we go to marshall the payload because we expect a string from `delivered_amount` field.

## Solution
Create an object to marshall and add `omitempty` flag to the struct.

```
{
  "hash": "1D849E3A0041357EE373C7E17C9564F890047475492D9530B5F20A3BD6D95822",
  "ledger_index": 49841027,
  "date": "2019-09-06T01:48:32+00:00",
  "tx": {
    "TransactionType": "Payment",
    "Flags": 2147942400,
    "Sequence": 292765,
    "LastLedgerSequence": 49841035,
    "Amount": {
      "value": "100000",
      "currency": "ETH",
      "issuer": "rJavT3eWaX9FubZFHtCvymJ6ZhSgJdMyNx"
    },
    "Fee": "162",
    "SendMax": "100000000000",
    "Account": "r4NT6UfELQyoS689VLye22B3SfgvpM3nHY",
    "Destination": "rJavT3eWaX9FubZFHtCvymJ6ZhSgJdMyNx"
  },
  "meta": {
    "TransactionIndex": 16,
    "DeliveredAmount": {
      "value": "533.92",
      "currency": "ETH",
      "issuer": "rJavT3eWaX9FubZFHtCvymJ6ZhSgJdMyNx"
    },
    "TransactionResult": "tesSUCCESS",
    "delivered_amount": {
      "value": "533.92",
      "currency": "ETH",
      "issuer": "rJavT3eWaX9FubZFHtCvymJ6ZhSgJdMyNx"
    }
  }
}
```

### DIFF
--- a/platform/ripple/model.go
+++ b/platform/ripple/model.go
@@ -33,7 +33,7 @@ type Payment struct {
 }
 
 type Meta struct {
-	DeliveredAmount string `json:"delivered_amount"`
+	DeliveredAmount string `json:"delivered_amount,omitempty"`
 }
 
 type LedgerResponse struct {
--- a/platform/ripple/model.go
+++ b/platform/ripple/model.go
@@ -33,7 +33,14 @@ type Payment struct {
 }
 
 type Meta struct {
-	DeliveredAmount string `json:"delivered_amount,omitempty"`
+	DeliveredAmount    string          `json:"delivered_amount,omitempty"`
+	DeliveredAmountObj DeliveredAmount `json:"delivered_amount,omitempty"`
+}
+
+type DeliveredAmount struct {
+	Value    string `json:"value"`
+	Currency string `json:"currency"`
+	Issuer   string `json:"issuer"`
 }
 
 type LedgerResponse struct {


Pantani:
> What? I did not know you can do that! Very awesome
> 
> We gonna show only XRP in the list correct?

No, we going return all transactions. But its simple to filter only the XRP transactions.
There is another type of transactions we can avoid, to create a destination tag, we need to do a transaction, we always show this type of transaction as well. This is a list of transactions type:

https://xrpl.org/transaction-types.html

vikmeup:
List of XRP Transactions: `http://blockatlas.trustwalletapp.com/v2/ripple/transactions/:address`
List of transaction for a specific token: `http://blockatlas.trustwalletapp.com/v2/ripple/transactions/:address?token=:token_id` 

This is a current flow, so in this case first API should only return xrp transactions, otherwise it would break iOS/Android clients

Pantani:
But in this case, we are using the block API to get all transactions from a block with the Blockatlas observer.

vikmeup:
For BlockAPI you can parse all transactions, but notify only for XRP, because we don't support any tokens on xrp

Pantani:
done!

vikmeup:
@madcake can you take a look at this one too

madcake:
How we can change headers? Is it work for other rpc?

Pantani:
I did some tests here, but we need to deploy and monitoring. the first 2 lines its means we gonna accept all headers and encodings, and the third can be a problem, but if we using JSON-RPC for all calls, I think don't have problem

```	
req.Header.Add("Accept-Encoding", "*")
req.Header.Add("Accept", "*/*")
req.Header.Add("Content-Type","application/json; charset=utf-8")
```

PR #357, opened 2019-09-13
TRON: improve transactions handling

kolya182:
Close https://github.com/trustwallet/backend/issues/151

Add check if address has any transactions before normalizing them
Increase max limit on returned trx, small lite caused some addresses not displayed transactions, but limit returned by 25 max

### DIFF
--- a/platform/tron/api.go
+++ b/platform/tron/api.go
@@ -25,17 +25,24 @@ func (p *Platform) Coin() coin.Coin {
 }
 
 func (p *Platform) GetTxsByAddress(address string) (blockatlas.TxPage, error) {
-	srcTxs, err := p.client.GetTxsOfAddress(address, "")
+	Txs, err := p.client.GetTxsOfAddress(address, "")
 	if err != nil {
 		return nil, err
 	}
 
+	if len(Txs) == 0 {
+		return nil, err
+	}
+
 	var txs []blockatlas.Tx
-	for _, srcTx := range srcTxs {
-		tx, ok := Normalize(&srcTx)
-		if ok {
-			txs = append(txs, tx)
+	for _, srcTx := range Txs {
+		if len(txs) <= blockatlas.TxPerPage {
+			tx, ok := Normalize(&srcTx)
+			if ok {
+				txs = append(txs, tx)
+			}
 		}
+
 	}
 
 	return txs, nil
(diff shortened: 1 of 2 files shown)


platform/tron/api.go
```diff
@@ -25,17 +25,24 @@ func (p *Platform) Coin() coin.Coin {
 }
 
 func (p *Platform) GetTxsByAddress(address string) (blockatlas.TxPage, error) {
-	srcTxs, err := p.client.GetTxsOfAddress(address, "")
+	Txs, err := p.client.GetTxsOfAddress(address, "")
 	if err != nil {
 		return nil, err
 	}
 
+	if len(Txs) == 0 {
```
vikmeup: why not do `if err != nil && len(Txs) == 0` ?

platform/tron/api.go
```diff
@@ -47,11 +54,20 @@ func (p *Platform) GetTokenTxsByAddress(address, token string) (blockatlas.TxPag
 		return nil, err
 	}
 
+	if len(tokenTxs) == 0 {
+		return nil, err
+	}
+
 	var tokenInfo AssetInfo
 	info, err := p.client.GetTokenInfo(token)
 	if err != nil {
```
vikmeup: can you add `len(info.Data) == 0` here too?

platform/tron/api.go
```diff
@@ -25,17 +25,24 @@ func (p *Platform) Coin() coin.Coin {
...
+		return nil, err
+	}
+
 	var txs []blockatlas.Tx
-	for _, srcTx := range srcTxs {
-		tx, ok := Normalize(&srcTx)
-		if ok {
-			txs = append(txs, tx)
+	for _, srcTx := range Txs {
+		if len(txs) <= blockatlas.TxPerPage {
```
vikmeup: can we just return all for now? not all of them might be normalized. we should not have this logic on each provider, generic one on not, so let's not change this

platform/tron/api.go
kolya182: https://github.com/trustwallet/blockatlas/pull/357/commits/f47a6982c4bc5af56d8b5fe366caf2e55eff9c15

platform/tron/api.go
kolya182: https://github.com/trustwallet/blockatlas/pull/357/commits/f47a6982c4bc5af56d8b5fe366caf2e55eff9c15

platform/tron/api.go
kolya182: Let's do in separate PR

vikmeup: can you remove this change? otherwise lgtm,

PR #363, opened 2019-09-16
create collectionId field

Pantani:
create `CollectionId` field to normalize collections API


collectibles.go
```diff
@@ -1,6 +1,7 @@
 package blockatlas
 
 type Collection struct {
+	CollectionId    string `json:"collection_id"`
```
vikmeup: can we just call id `id` ? it's already returning collection model

madcake: In client used collection_id. In next api version good idea use id.

vikmeup: which client? `v2/ethereum/collections/0x5574Cd97432cEd0D7Caf58ac3c4fEDB2061C98fB` does not have `collection_id `

madcake: I looked /collectibles/category.

vikmeup: `id` should be good to go then!

PR #364, opened 2019-09-17
Handling errors gracefully

Pantani:
# Headline

## Motivation
- Easy to build informative error messages;
- Errors easy to understand for users;
- Having errors more helpful for programmers diagnostics;

## Problem
Some errors it's hard to tracking and interpreter because some information is missing.

## Solution
An error in Go is any implementing interface with an Error() string method. We overwrite the error object by our error struct:

```
type Error struct {
	Err   error
	Type  Type
	meta  map[string]interface{}
	stack []string
}
```

To be easier the error construction, the package provides a function named E, which is short and easy to type:

`func E(args ...interface{}) *Error`

E.g.:
- just error:
`errors.E(err)`

- error with message:
`errors.E(err, "new message to append")`

- error with type:
`errors.E(err, errors.TypePlatformReques)`

- error with type and message:
`errors.E(err, errors.TypePlatformReques, "new message to append")`

- error with type and meta:
```
errors.E(err, errors.TypePlatformRequest, errors.Params{
			"coin":   "Ethereum",
			"method": "CurrentBlockNumber",
		})
```

- error with meta
```
errors.E(err, errors.Params{
			"coin":   "Ethereum",
			"method": "CurrentBlockNumber",
		})
```

- error with type and meta
```
errors.E(err, errors.TypePlatformRequest, errors.Params{
			"coin":   "Ethereum",
			"method": "CurrentBlockNumber",
		})
```

- error with type, message and meta:
```
errors.E(err, errors.TypePlatformRequest, "new message to append", errors.Params{
			"coin":   "Ethereum",
			"method": "CurrentBlockNumber",
		})
```

### Types

```
const (
	TypeNone Type = iota
	TypePlatformUnmarshal
	TypePlatformNormalize
	TypePlatformUnknown
	TypePlatformRequest
	TypePlatformClient
	TypePlatformError
	TypePlatformApi
	TypeLoadConfig
	TypeLoadCoins
	TypeObserver
	TypeStorage
	TypeAssets
	TypeUtil
	TypeCmd
	TypeUnknown
)
```

## Result
- Now the error is a simple JSON with all information and context:
`{"caller":"/Users/danilopantani/Desktop/Go/src/github.com/trustwallet/blockatlas/cmd/api/run.go:61","error":"listen tcp: lookup tcp/failed_bind: nodename nor servname provided, or not known","type":"Platform Request Error"}`

[example](https://github.com/trustwallet/blockatlas/pull/364/files#diff-5098d6c9f8c575625d507e8257bc2f88R76)
Screen Shot 2019-09-17 at 00 34 48

### DIFF
--- a/pkg/errors/errorType.go
+++ b/pkg/errors/errorType.go
@@ -1,0 +1,62 @@
+package errors
+
+import (
+	"fmt"
+)
+
+type Type uint16
+
+const (
+	TypeNone Type = iota
+	TypePlatformUnmarshal
+	TypePlatformNormalize
+	TypePlatformUnknown
+	TypePlatformRequest
+	TypePlatformApi
+	TypeStorageSave
+	TypeStorageGet
+	TypeLoadConfig
+	TypeLoadCoins
+	TypeObserver
+	TypeAssets
+	TypeUtil
+	TypeCmd
+	TypeUnknown
+)
+
+func (e Type) String() string {
+	switch e {
+	case TypeNone:
+		return ""
+	case TypePlatformRequest:
+		return "Platform Request Error"
+	case TypePlatformUnmarshal:
+		return "Platform Unmarshal Error"
+	case TypePlatformApi:
+		return "Platform API Error"
+	case TypePlatformNormalize:
+		return "Platform Normalize Error"
+	case TypePlatformUnknown:
+		return "Platform Unknown Error"
+	case TypeObserver:
+		return "Observer Error"
+	case TypeStorageSave:
+		return "Storage Save Error"
+	case TypeStorageGet:
+		return "Storage Get Error"
+	case TypeLoadConfig:
+		return "Load Config Error"
+	case TypeLoadCoins:
+		return "Load Coins Error"
+	case TypeAssets:
+		return "Assets Error"
+	case TypeUtil:
+		return "Util Error"
+	case TypeCmd:
+		return "Cmd Error"
+	case TypeUnknown:
+		return "Unknown Error"
+	default:
+		return fmt.Sprintf("Error: %d", int(e))
+	}
+}
--- a/pkg/errors/errorType.go
+++ b/pkg/errors/errorType.go
@@ -12,7 +12,9 @@ const (
 	TypePlatformNormalize
 	TypePlatformUnknown
 	TypePlatformRequest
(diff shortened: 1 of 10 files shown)


pkg/errors/errorType.go
```diff
@@ -0,0 +1,62 @@
+package errors
```
madcake: Will be problems with the file name? Could we move it to errros.go?

platform/ethereum/client.go
```diff
@@ -72,14 +73,20 @@ func (c *Client) CurrentBlockNumber() (num int64, err error) {
 	path := fmt.Sprintf("%s/node_info", c.BaseURL)
 	res, err := http.Get(path)
 	if err != nil {
-		return num, err
+		return num, errors.E(err, errors.TypePlatformRequest, errors.Params{
+			"coin":   "Ethereum",
```
vikmeup: can we handle this error on the higher level? so it would log all errors for all providers

pkg/errors/errorType.go
Pantani: we don't need more the built-in errors package. Our package overwrites the error. But we can use both with the same name without problems

platform/ethereum/client.go
Pantani: We don't go to handle now. We can use this package to put more information inside the error. This error doesn't go to be logged, now. We are going to handle this in higher live with the logger package

pkg/logger/logger.go
```diff
@@ -18,65 +18,67 @@ func InitLogger() {
 	}
 }
 
-type Msg struct {
-	Message string
-	Params  map[string]interface{}
+type message struct {
```
madcake: Why is this structure private?

Pantani: because we don't need to access this variable from external packages

Pantani:
> Please update readme
> 
> What if sentry credentials not specified will error be printed console ?
Done


Yes!
`17 Sep 2019 20:31:23.462189 <190>1 2019-09-17T23:31:20.595046+00:00 app web.1 - - time="2019-09-17T23:31:20Z" level=error msg="{\"error\":\"InitSentry failed\",\"meta\":\"{}\"}"`

PR #366, opened 2019-09-17
Change the built-in errors to our error structure

Pantani:
# Headline
Change the built-in errors to our error structure.

## Problem
- Some errors it's hard to tracking and interpreter because some information is missing;
- We send all errors to Sentry when we call the error log method, confusing the Sentry and grouping different errors in the same issue;

## Solution
Change all built-in errors in the root of the cause to our error structure with more information, to send to the Sentry in the background. And we can log the errors only in the higher-level of the code.

## Changes
- Change all built-in errors in the root cause;
- Add logs add high-level of the code;
- Now we send to Sentry when we create an error;
- Remove useless log;
- Small fixes;

_**Sorry by the extensive PR. I need to change all errors at the same time to don't lose the trackability in the Sentry.**_

https://www.youtube.com/watch?v=4WIhhzTTd0Y


pkg/errors/errors.go
```diff
@@ -123,6 +123,6 @@ func E(args ...interface{}) *Error {
 		msg := strings.Join(message[:], ": ")
 		e.Err = errors.New(msg)
 	}
-
+	SendError(e)
```
Pantani: now we are going to send to Sentry here

pkg/logger/error.go
```diff
@@ -16,7 +16,6 @@ func Error(args ...interface{}) {
 	}
 	e := getError(args...)
 	log.WithFields(e.params).Error(e.err)
-	errors.SendError(e.err)
```
Pantani: and we are going to stop to send to Sentry from the log package

kolya182:
`Log.Error`, `Log.Info` feels more natural to type. `errors.E` is less intuitive` IMHO

Pantani:
@kolya182 `logger.Error` and `logger.Info` do different things that `errors.E`. The method `errors.E` is similar to `fmt.Errorf`  or `errors.New`. We are going to create a new error, but we can put more information inside this error in the root of the cause. And we can use this method to send to Sentry as well. And `logger.Error` now is only to log the error in the Logentries

PR #367, opened 2019-09-18
Auto generate the coin list

Pantani:
# Headline
Now we can generate a coin list and the coin enum from the `gen.go` file


coin/coins.go
```diff
@@ -0,0 +1,482 @@
...
+	XZC = 136
+	BCH = 145
+	RVN = 175
+	QTUM = 2301
+	ZEL = 19167
+	DCR = 42
+)
+
+var Coins = map[uint]Coin{
+	60: {
```
vikmeup: can we change `60` as `ID` to be `ethereum` as handle?

platform/nebulas/api_test.go
```diff
@@ -66,6 +68,8 @@ func TestNormalize(t *testing.T) {
 	}
 
 	if !bytes.Equal(resJSON, dstJSON) {
+		println(string(resJSON))
+		println(string(dstJSON))
```
vikmeup: ca we use `"github.com/stretchr/testify/assert"` to assert easier?

Pantani: Yes, I did this PR to use the auto-gen coin file, but some tests are broke because there are missing parameters, I just fix simple things.

PR #371, opened 2019-09-20
add unit tests to auto-generated coin file

Pantani:


Pantani:
> Can you also add some real test cases? Like ethereum ID is equal to 60 and etc.
I think this type of test doesn't make sense for an auto-generated file. Because if we going to stay this type of things hardcode, we don't need an auto-generated file

> Can we rename Title to Name?
Done

vikmeup:
But how do you make sure that coin.ETH even has correct data? Or if you do coin.Ethereum().Name even return correct name? For that reason one unit test to confirm that coin 

Check Wallet Core tests cases that we do for this.

PR #372, opened 2019-09-20
Send http metrics to Prometheus

Pantani:
# Headline
Send HTTP metrics to Prometheus using expvar's.

## Problem
Today we don't have metrics about all request made to the application and all requests made from the application. 

## Solution
Now we can send these metrics to an [expvar](https://golang.org/pkg/expvar/) and pooling this vars using the Prometheus to show inside Grafana

[Example - Stage Environment](https://blockatlas-trust-staging.herokuapp.com/metrics)


Pantani:
> Please update readme, give overview how to setup

We don't need to do anything inside the Block Atlas. Expvars only expose the metrics, the overview going to be described inside the monitoring tool repository.

config.yml
```diff
@@ -6,6 +6,9 @@ gin:
   # If set, HTTP Forwarded headers will be respected
   reverse_proxy: false
 
+api:
```
vikmeup: should this be mertrics api token then?

PR #375, opened 2019-09-23
add default client

Pantani:
Add default client for APIs with 10 seconds timeout

### DIFF
--- a/client.go
+++ b/client.go
@@ -7,11 +7,16 @@ import (
 	"net/http"
 	"net/url"
 	"strings"
+	"time"
 )
 
 type Request struct {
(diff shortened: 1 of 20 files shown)


client.go
```diff
@@ -7,13 +7,18 @@ import (
...
+	"time"
 )
 
 type Request struct {
 	HttpClient   *http.Client
 	ErrorHandler func(res *http.Response, uri string) error
 }
 
+var DefaultClient = &http.Client{
+	Timeout: time.Second * 10,
```
vikmeup: let's do 15 just in case

Pantani: great!

Issue #273, opened 2019-08-25
Refactor old zilliqa client and use the general one


hewigovens:
See `blockatlas/client.go`, need to add http headers support

Pantani:
#420

PR #380, opened 2019-09-23
Improve Blockatlas client

Pantani:
# Headline
Improve client struct and remove useless code

## Problem
We always need to pass an URL to do the requests. and the GET method it's verbose to call.

## Solution
- Pass the Base Url to the Request struct
- We can do a GET request with `c.GET`, and no more `c.request.GET`;

### DIFF
--- a/platform/binance/dex.go
+++ b/platform/binance/dex.go
@@ -1,0 +1,36 @@
+package binance
+
+import (
+	"fmt"
+	"github.com/trustwallet/blockatlas/pkg/blockatlas"
+	"net/url"
+)
+
+// TODO Headers + rate limiting
+
+type DexClient struct {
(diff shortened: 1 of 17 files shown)


platform/binance/dex.go
```diff
@@ -0,0 +1,36 @@
...
+	"net/url"
+)
+
+// TODO Headers + rate limiting
+
+type DexClient struct {
+	blockatlas.Request
+}
+
+func ClientDexInit(baseUrl string) DexClient {
```
vikmeup: maybe call file `dex-client.go`? or whichever format is common

Pantani:
> can we also do this: #273? need to set http header

Yes, refactor other platforms is the next work

Issue #390, opened 2019-09-26
/delegations endpoint has missing information about the validator


vikmeup:
`/v2/cosmos/staking/delegations/cosmos1cxehfdhfm96ljpktdxsj0k6xp9gtuheghwgqug` endpoint does not have information about the delegator as `/v2/cosmos/staking/validators`

Expected response: 

```json
[
   {
      "delegator":{
         "id":"cosmosvaloper1qwl879nx9t6kef4supyazayf7vjhennyh568ys",
         "status":true,
         "info":{
            "name":"Certus One",
            "description":"Stake and earn rewards with the most secure and stable validator. Winner of the Game of Stakes. Operated by Certus One Inc. By delegating, you confirm that you are aware of the risk of slashing and that Certus One Inc is not liable for any potential damages to your investment.",
            "image":"https://raw.githubusercontent.com/trustwallet/assets/master/blockchains/cosmos/validators/assets/cosmosvaloper1qwl879nx9t6kef4supyazayf7vjhennyh568ys/logo.png",
            "website":"https://certus.one"
         },
         "reward":{
            "annual":9.282023135552596
         },
         "locktime":1814400,
         "minimum_amount":"0"
      },
      "coin":{
         "coin":118,
         "symbol":"ATOM",
         "name":"Cosmos ATOM",
         "decimals":6
      },
      "value":"3000000",
      "status":"active"
   }
]
```

Pantani:
#395

PR #383, opened 2019-09-25
Add delegators models

vikmeup:
Implement: https://github.com/trustwallet/blockatlas/issues/382

- [ ] `ExternalCoin` maybe rename ?
- [x]  Conform Tron blockchain to `StakeAPI `
- [x]  Conform Tezos blockchain to `StakeAPI ` https://github.com/trustwallet/blockatlas/issues/386

PR #395, opened 2019-09-27
Add validator info for delegators

Pantani:
# Headline
Add validator info for delegators endpoint.

## Problem
/delegations endpoint has missing information about the validator.

[Issue 390](https://github.com/trustwallet/blockatlas/issues/390)

### DIFF
--- a/platform/cosmos/api.go
+++ b/platform/cosmos/api.go
@@ -2,6 +2,7 @@ package cosmos
 
 import (
 	"github.com/trustwallet/blockatlas/pkg/blockatlas"
+	"github.com/trustwallet/blockatlas/pkg/logger"
 	"sort"
 	"strconv"
 	"time"
@@ -9,6 +10,8 @@ import (
 	"github.com/spf13/viper"
 	"github.com/trustwallet/blockatlas/coin"
 	"github.com/trustwallet/blockatlas/util"
+
+	services "github.com/trustwallet/blockatlas/services/assets"
 )
 
 type Platform struct {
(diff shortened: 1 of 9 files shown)


platform/cosmos/api.go
```diff
@@ -101,19 +104,30 @@ func (p *Platform) GetDelegations(address string) (blockatlas.DelegationsPage, e
...
 	return results, nil
 }
 
-func NormalizeDelegations(delegations []Delegation) []blockatlas.Delegation {
+func NormalizeDelegations(delegations []Delegation, validators blockatlas.ValidatorMap) []blockatlas.Delegation {
 	c := coin.Cosmos()
 	results := make([]blockatlas.Delegation, 0)
 	for _, v := range delegations {
+		validator, ok := validators[v.ValidatorAddress]
+		if !ok {
```
vikmeup: Should we add continue here is not found?

Pantani: I don't know, this info is crucial to show?

Pantani:
> Maybe we should just merge validators on higher levels than do it for each platform?
> 
> Similar example how we merge github list with platform list

If we merge just the higher level, the delegations for some platforms in production going to break

vikmeup:
Why wound it break? 

We should make it reusable, it does not make sense to merge that on every platform

Pantani:
> Why wound it break?
> 
> We should make it reusable, it does not make sense to merge that on every platform

because of the model changes
https://github.com/trustwallet/blockatlas/pull/395/files#diff-289e7eeff75e55aa7f73ff0d65354063R36

vikmeup:
@Pantani What stopping you from creating temp model before merging? 

I don’t see any issues by improving model structure, otherwise you add complexity for each platform to implement staking protocol api

Pantani:
> @Pantani What stopping you from creating temp model before merging?
> 
> I don’t see any issues by improving model structure, otherwise you add complexity for each platform to implement staking protocol api

I did similar the other PR
https://github.com/trustwallet/blockatlas/pull/383/files

I going to split, no problem

PR #398, opened 2019-09-28
Cosmos delegator info

Pantani:
need merge #397

### DIFF
--- a/platform/cosmos/api.go
+++ b/platform/cosmos/api.go
@@ -1,14 +1,15 @@
 package cosmos
 
 import (
+	"github.com/spf13/viper"
+	"github.com/trustwallet/blockatlas/coin"
 	"github.com/trustwallet/blockatlas/pkg/blockatlas"
+	"github.com/trustwallet/blockatlas/pkg/logger"
+	services "github.com/trustwallet/blockatlas/services/assets"
+	"github.com/trustwallet/blockatlas/util"
 	"sort"
 	"strconv"
 	"time"
-
-	"github.com/spf13/viper"
-	"github.com/trustwallet/blockatlas/coin"
-	"github.com/trustwallet/blockatlas/util"
 )
 
 type Platform struct {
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -24,4 +24,5 @@ require (
 	github.com/ybbus/jsonrpc v2.1.2+incompatible
 	golang.org/x/crypto v0.0.0-20190701094942-4def268fd1a4
 	golang.org/x/net v0.0.0-20190620200207-3b0461eec859 // indirect
+	gopkg.in/yaml.v2 v2.2.2
```
vikmeup: is this something new?

platform/cosmos/api.go
```diff
@@ -101,19 +101,30 @@ func (p *Platform) GetDelegations(address string) (blockatlas.DelegationsPage, e
...
 }
 
-func NormalizeDelegations(delegations []Delegation) []blockatlas.Delegation {
+func NormalizeDelegations(delegations []Delegation, validators blockatlas.ValidatorMap) []blockatlas.Delegation {
 	c := coin.Cosmos()
 	results := make([]blockatlas.Delegation, 0)
 	for _, v := range delegations {
+		validator, ok := validators[v.ValidatorAddress]
+		if !ok {
+			logger.Error("Validator not found", validator)
```
vikmeup: do you need to continue?

Pantani: You know more than me, can you help me with this question? Do we need to continue? I think we don't have a problem. We can show just the JSON below in this case.

```
[
   {
      "delegator":{
         "id":"cosmosvaloper1qwl879nx9t6kef4supyazayf7vjhennyh568ys",
      },
      "coin":{
         "coin":118,
         "symbol":"ATOM",
         "name":"Cosmos ATOM",
         "decimals":6
      },
      "value":"3000000",
      "status":"active"
   }
]
```

vikmeup: i don't think it make sense to return this way. if validated not found with information we should not present it here.

vikmeup: we should only present validators that provided information in assets repo

Pantani: great. I going to fix this. Thks

PR #400, opened 2019-09-28
Add algorand transactions

vikmeup:
Algorand promised to deliver an endpoint that will have ability to see timestamp of each transaction, but for now we need to have a hack to get block for each transaction and then merge that value.

Currently it only fetches last 6 transactions, which is enough - once Algorand deliver new endpoint we will change this logic.

SHITCODE DETECTED IN THIS PR.  💥

### DIFF
--- a/platform/algorand/api.go
+++ b/platform/algorand/api.go
@@ -1,17 +1,19 @@
 package algorand
 
 import (
+	"github.com/spf13/viper"
 	//"github.com/spf13/viper"
 	"github.com/trustwallet/blockatlas/coin"
 	"github.com/trustwallet/blockatlas/pkg/blockatlas"
+	"strconv"
 )
 
 type Platform struct {
-	//client Client
+	client Client
 }
 
 func (p *Platform) Init() error {
-	//p.client = InitClient(viper.GetString("algorand.api"))
+	p.client = InitClient(viper.GetString("algorand.api"))
 	return nil
 }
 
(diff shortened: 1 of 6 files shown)


platform/algorand/api_test.go
```diff
@@ -1,78 +1,76 @@
...
+			Symbol:   "ALGO",
+			Decimals: 6,
+		},
+	},
+	nil,
+	nil,
+}
+
+func TestNormalize(t *testing.T) {
+	a := assert.New(t)
```
Pantani: Do we need to instantiate this?

platform/algorand/api_test.go
```diff
@@ -1,78 +1,76 @@
...
+
+	var act TransactionsResponse
+	a.NoError(json.Unmarshal([]byte(transfer), &act))
+	a.Equal(1, len(act.Transactions))
+
+	for i, v := range act.Transactions {
+		tx, ok := Normalize(v)
+		assert.True(t, ok)
+		assert.Equal(t, expected[i], &tx)
```
Pantani: why you use global here and local above?

platform/algorand/api.go
```diff
@@ -1,83 +1,82 @@
...
+		return nil, err
+	}
+
+	return &blockatlas.Block{
+		Number: num,
+		Txs:    NormalizeTxs(txs),
+	}, nil
+}
 
 func (p *Platform) GetTxsByAddress(address string) (page blockatlas.TxPage, err error) {
```
Pantani:
```suggestion
func (p *Platform) GetTxsByAddress(address string) (blockatlas.TxPage, error) {
```

platform/algorand/api.go
```diff
@@ -1,83 +1,82 @@
...
+	for _, tx := range txs {
+		if normalized, ok := Normalize(tx); ok {
+			result = append(result, normalized)
+		}
+	}
+
+	return result
+}
+
+func Normalize(tx Transaction) (result blockatlas.Tx, ok bool) {
```
Pantani:
```suggestion
func Normalize(tx Transaction) (blockatlas.Tx, bool) {
```

platform/algorand/api.go
```diff
@@ -1,83 +1,82 @@
...
+
+	if tx.Type != TransactionTypePay {
+		return result, false
```
Pantani:
```suggestion
		return blockatlas.Tx{}, false
```

platform/algorand/api.go
```diff
@@ -1,83 +1,82 @@
...
+	}
+
+	return blockatlas.Tx{
+		ID:     tx.Hash,
+		Coin:   coin.ALGO,
+		From:   tx.From,
+		To:     tx.Payment.To,
+		Fee:    blockatlas.Amount(strconv.Itoa(int(tx.Fee))),
+		Date:   int64(tx.Timestamp),
+		Block:  uint64(tx.Round),
```
Pantani:
```suggestion
		Block:  tx.Round,
```

platform/algorand/api_test.go
```diff
@@ -1,78 +1,76 @@
...
+)
+
+var expected = []*blockatlas.Tx{
+	{
+		ID:     "C2LK3CGBPIGERLPFUXE6INSBJGHOXU7YZMEGELWMVSBASFJYOOQQ",
+		Coin:   coin.ALGO,
+		From:   "5TSQNIL54GB545B3WLC6OVH653SHAELMHU6MSVNGTUNMOEHAMWG7EC3AA4",
+		To:     "4EZFQABCVQTHQCK3HQBIYGC4NV2VM42FZHEFTVH77ROG4ZGREC6Y7V5T2U",
+		Fee:    blockatlas.Amount("1000"),
+		Date:   int64(0),
```
Pantani:
```suggestion
		Date:   0,
```

platform/algorand/client.go
```diff
@@ -1,54 +1,88 @@
...
+	}
+}
+
+func (c *Client) GetLatestBlock() (int64, error) {
+	var status Status
+	err := c.Request.Get(&status, "v1/status", nil)
+	if err != nil {
+		return 0, err
+	}
+	return int64(status.LastRound), nil
```
Pantani:
```suggestion
	return status.LastRound, nil
```

platform/algorand/client.go
```diff
@@ -1,54 +1,88 @@
...
-//	return response.Transactions, err
-//}
+import (
+	"fmt"
+	"github.com/trustwallet/blockatlas/pkg/blockatlas"
+	"github.com/trustwallet/blockatlas/util"
+)
+
+type Client struct {
+	Request blockatlas.Request
```
Pantani:
```suggestion
	blockatlas.Request
```

platform/algorand/client.go
```diff
@@ -1,54 +1,88 @@
...
+			HttpClient:   blockatlas.DefaultClient,
+			ErrorHandler: blockatlas.DefaultErrorHandler,
+			BaseUrl:      URL,
+		},
+	}
+}
+
+func (c *Client) GetLatestBlock() (int64, error) {
+	var status Status
+	err := c.Request.Get(&status, "v1/status", nil)
```
Pantani:
```suggestion
	err := c.Get(&status, "v1/status", nil)
```

platform/algorand/client.go
```diff
@@ -1,54 +1,88 @@
...
+	if err != nil {
+		return 0, err
+	}
+	return int64(status.LastRound), nil
+}
+
+func (c *Client) GetBlock(number int64) (BlockResponse, error) {
+	path := fmt.Sprintf("v1/block/%d", number)
+	var resp BlockResponse
+	err := c.Request.Get(&resp, path, nil)
```
Pantani:
```suggestion
	err := c.Get(&resp, path, nil)
```

platform/algorand/client.go
```diff
@@ -1,54 +1,88 @@
...
+func (c *Client) GetTxsInBlock(number int64) ([]Transaction, error) {
+	block, err := c.GetBlock(number)
+	return block.Transactions.Transactions, err
+}
+
+func (c *Client) GetTxsOfAddress(address string) ([]Transaction, error) {
+	var response TransactionsResponse
+	path := fmt.Sprintf("v1/account/%s/transactions", address)
+
+	err := c.Request.Get(&response, path, nil)
```
Pantani:
```suggestion
	err := c.Get(&response, path, nil)
```

util/number.go
```diff
@@ -0,0 +1,15 @@
+package util
+
+func Min(x, y int) int {
```
Pantani: add unit tests

util/number.go
```diff
@@ -0,0 +1,15 @@
...
+	if x < y {
+		return x
+	}
+	return y
+}
+
+func Max(x, y int) int {
```
Pantani: add unit tests

platform/algorand/client.go
```diff
@@ -1,54 +1,88 @@
...
+	//FIXME. Currently fetching the last 8 transactions and get 6 blocks for each to retrieve timestamp.
+	//Algorand team promised to provide endpoint soon that will contain timestamp value inside TransactionsResponse response
+	//Get latest 8 transactions, which is enough until new endpoint fixes it.
+	txs := response.Transactions[:util.Min(6, len(response.Transactions))]
+
+	for _, t := range txs {
+		block, err := c.GetBlock(int64(t.Round))
+		if err != nil {
+			continue
+		} else {
```
Pantani: you don't need this else

platform/algorand/api.go
vikmeup: result here is `blockatlas.Tx{}` already

platform/algorand/api.go
vikmeup: if there is early return, it's easier ti have variable specified, then you can return it as we do on line 63

vikmeup: is there a standard for this? I have seen other places having same structure

platform/algorand/client.go
vikmeup: `LastRound` is `uint64`  and this function requires int, we need to have this in other places

platform/algorand/api.go
Pantani: you going to reserve memory space for this variable but you only use if an error occurs. Go is pretty fast, we don’t need to worry about this. But we can keep the good practices

Issue #384, opened 2019-09-25
Implement endpoint to fetch delegations for an array of addresses


vikmeup:
Currently each blockchain will implement `v2/cosmos/staking/delegations/cosmos1cxehfdhfm96ljpktdxsj0k6xp9gtuheghwgqug` that returns list of delegations active and pending.

To optimize even further it would be great to return list of delegations for all chains.
An example:
POST `v2/staking/delegations`

payload:

```json
[
    {
       "coin": 118,
       "address":  "123",
    },
   {
       "coin": 118,
       "address":  "321",
    }
]
```

Response:
```json
{
   "docs":[
      {
         "coin":118,
         "address":"123",
         "delegations":[
            {
               "delegator":"cosmosvaloper1qwl879nx9t6kef4supyazayf7vjhennyh568ys",
               "coin":{
                  "coin":118,
                  "symbol":"ATOM",
                  "name":"Cosmos ATOM",
                  "decimals":6
               },
               "value":"1000000",
               "status":"active"
            },
            {
               "delegator":"cosmosvaloper1qwl879nx9t6kef4supyazayf7vjhennyh568ys",
               "coin":{
                  "coin":118,
                  "symbol":"ATOM",
                  "name":"Cosmos ATOM",
                  "decimals":6
               },
               "value":"1000000",
               "status":"active"
            }
         ]
      },
      {
         "coin":118,
         "address":"321",
         "delegations":[
            {
               "delegator":"cosmosvaloper1qwl879nx9t6kef4supyazayf7vjhennyh568ys",
               "coin":{
                  "coin":118,
                  "symbol":"ATOM",
                  "name":"Cosmos ATOM",
                  "decimals":6
               },
               "value":"1000000",
               "status":"active"
            },
            {
               "delegator":"cosmosvaloper1qwl879nx9t6kef4supyazayf7vjhennyh568ys",
               "coin":{
                  "coin":118,
                  "symbol":"ATOM",
                  "name":"Cosmos ATOM",
                  "decimals":6
               },
               "value":"1000000",
               "status":"active"
            }
         ]
      }
   ]
}
```

Pantani:
#402

PR #402, opened 2019-09-29
add batch call for delegations

Pantani:
[Issue 384](https://github.com/trustwallet/blockatlas/issues/384)


api/batch.go
```diff
@@ -0,0 +1,48 @@
...
+
+func makeStakingDelegationsBatchRoute(router gin.IRouter) {
+	router.POST("/staking/delegations/", func(c *gin.Context) {
+		var reqs AddressesRequest
+		if c.BindJSON(&reqs) != nil {
+			return
+		}
+
+		if len(reqs) == 0 {
+			RenderSuccess(c, EmptyResponse)
```
vikmeup: I don’t think we need to check req, if it’s 0, it will return later anyway

PR #403, opened 2019-09-30
Provide swagger docs

Pantani:
Provide swagger docs at path `/swagger/index.html`


docs/swagger.yaml
```diff
@@ -0,0 +1,578 @@
+basePath: '{{.BasePath}}'
```
hewigovens: maybe we only need one? swagger.json or swagger.yaml

Pantani: this is automatically generated by [gin-swagger](https://github.com/swaggo/gin-swagger), to keep only one we need to generate and delete the other. I think we can keep them both

Issue #342, opened 2019-09-12
Add caching layer for staking requests


vikmeup:
for requests as `cosmos/staking/validators` or `tezos/staking/validators`, `tron/staking/validators` we can add caching, as 5 minutes or so, since it does not change much, but currently we need to do an extra request to assets repo on github and then rpc.

Pantani:
#426

PR #424, opened 2019-10-02
Add caching layer to staking requests

vikmeup:
Fixes: https://github.com/trustwallet/blockatlas/issues/342


vikmeup:
@Pantani can you take a look, this solution work, but i don't like how it's currently done on each handlers level

PR #426, opened 2019-10-02
Create cache middleware

Pantani:
- Create cache middleware with [go-cache](https://github.com/patrickmn/go-cache);
- Cache for staking validator route;


api/cache.go
```diff
@@ -0,0 +1,123 @@
...
+			w.Status(),
+			w.Header(),
+			[]byte(data),
+		}
+		memoryCache.Set(w.key, val, w.expire)
+	}
+	return ret, err
+}
+
+func CacheMiddleware(minLength int, expiration time.Duration, handle gin.HandlerFunc) gin.HandlerFunc {
```
vikmeup: What is minLenght?

Pantani: I added a parameter minLength. It's used in case the API returns empty docs, like { "docs": [] }, we don't gonna cache that

vikmeup: I would suggest to remove `minLength` and make `validators()` to fail if it wasn't able to fetch data. in this way it would be very specific. 

Unless better way to handle it?

Pantani: it's better!!! done!!

PR #443, opened 2019-10-08
Storage package

Pantani:
Create a generic package to handle database 

- Create a new generic Redis package
- Add Postgres support


pkg/storage/sql/postgres.go
```diff
@@ -0,0 +1,63 @@
...
+		return storage.ErrNotFound
+	}
+	err := db.client.Save(value).Error
+	if err != nil {
+		return errors.E(err, storage.ErrNotUpdated)
+	}
+	return nil
+}
+
+func (db *PgSql) Add(value interface{}) error {
```
vikmeup: do you need to specify table name?  or how does it work?

pkg/storage/storage.go
```diff
@@ -0,0 +1,32 @@
...
+)
+
+var Redis redis.Redis
+var Postgres sql.PgSql
+
+func InitDatabases() {
+	err := Redis.Init(viper.GetString("observer.redis"))
+	if err != nil {
+		logger.Fatal(er.E(err), "Cannot connect to Redis")
+	}
```
vikmeup: can we potentially use just one database? because if one of the databases is down - blockatlas will fail

Pantani: I like Redis to storage the block numbers, is quickly and don't need a lot of processing
But we can keep only one database to don't make the infra more complex

pkg/storage/sql/postgres.go
Pantani: I did use Gorm, it's a Golang ORM. It automatically creates my databases based on the structs

https://gorm.io/

Pantani: I going start to implement and you going to see is easier to use

Pantani:
> Are we going to use MySQL / PQ at the same time?

only Postgres, but to change the SQL provides is easily just one line

Pantani:
> @Pantani do you use Pg as NoSql db?

We as use as SQL (Relational) db

kolya182:
@Pantani Could you justify use of PG next time when making such radical changes

Pantani:
> @Pantani Could you justify use of PG next time when making such radical changes

Yes.
This PR don't make changes, only introduce a new package to handle databases. The database changes going to be in another PR. We need a better database with queries to handle the xpub things. I going to explain better and create a readme in the next PR

PR #445, opened 2019-10-10
Add staking details to the root of delegations API

vikmeup:
New response in https://github.com/trustwallet/platform/issues/192

In order to present useful information for each delegations, we need to have details exposed

### DIFF
--- a/platform/cosmos/api.go
+++ b/platform/cosmos/api.go
@@ -91,6 +91,15 @@ func (p *Platform) GetValidators() (blockatlas.ValidatorPage, error) {
 	return results, nil
 }
 
+func (p *Platform) GetDetails() blockatlas.StakingDetails {
+	//TODO: Find a way to have a dynamic
+	return blockatlas.StakingDetails{
+		Reward:        blockatlas.StakingReward{Annual: 11},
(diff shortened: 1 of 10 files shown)


platform/cosmos/api.go
```diff
@@ -91,6 +91,15 @@ func (p *Platform) GetValidators() (blockatlas.ValidatorPage, error) {
 	return results, nil
 }
 
+func (p *Platform) GetDetails() blockatlas.StakingDetails {
+	//TODO: Find a way to have a dynamic
+	return blockatlas.StakingDetails{
+		Reward:        blockatlas.StakingReward{Annual: 11},
+		MinimumAmount: blockatlas.Amount("0"),
+		LockTime:      1814400,
```
dpereskokov: Lets do it dynamic? we can do like that:
curl -X GET "https://stargate.cosmos.network/staking/parameters" -H "accept: application/json"
Field: unbonding_time

vikmeup: It's dynamic for delegations API,

it's only static for blockchain details itself, not per delegation

dpereskokov: Why not to do it dynamic too?

PR #446, opened 2019-10-11
add batch transactions for SQL

Pantani:


pkg/storage/sql/sql.go
```diff
@@ -7,47 +7,81 @@ import (
...
 		return errors.E(err, util.ErrNotDeleted)
 	}
 	return nil
 }
+
+func (db *sql) DeleteMany(values ...interface{}) error {
+	return db.ToMany(db.Delete, values...)
+}
+
+func (db *sql) ToMany(handler Handler, values ...interface{}) error {
```
madcake: Could you named it is Batch***?

pkg/storage/sql/sql.go
```diff
@@ -7,47 +7,81 @@ import (
...
+func (db *sql) CreateOrUpdate(value interface{}) error {
+	if db.Client.NewRecord(value) {
+		return db.Add(value)
+	}
+	return db.Update(value)
+}
+
 func (db *sql) Update(value interface{}) error {
-	if db.client.NewRecord(value) {
+	if db.Client.NewRecord(value) {
```
madcake: Could you do this function as proxy for toMany?

Pantani: do you say to write this verification only one time?

I think not, because is some cases we need to verify exist and other cases if don't exist

pkg/storage/sql/sql.go
Pantani: done

PR #451, opened 2019-10-11
Change Redis to Postgres

Pantani:
## Headline
Use Postgres instead Redis as database


artall64:
Hi, is it any reason to change ? 
As I see it's a full replacement of `redis` not like adding one alternative of storage. Am I correct ?

Pantani:
> Hi, is it any reason to change ?
> As I see it's a full replacement of `redis` not like adding one alternative of storage. Am I correct ?

The Redis is not used well.
Redis doesn't do queries, and we start to create some redundancies, and the memory did start to increase a lot.

observer/storage/blockmap.go
```diff
@@ -5,7 +5,7 @@ import (
 )
 
 type BlockMap struct {
-	heights map[int]*Block
+	heights map[interface{}]*Block
 	lock    sync.RWMutex
```
vikmeup: what's the purpose to use interface instead of int?

Pantani: Because bitcoin id is 0, if I keep the coin with int in the database, I cannot use this field as a primary key because zero is null. So, we pass the int for the database model, but in the database going to store as a string. The GORM does the casting.

https://github.com/trustwallet/blockatlas/pull/451/files#diff-517d2699fdea1888a8a32a0462a21152R6

PR #467, opened 2019-10-16
Implement fetching balance for Tron

dpereskokov:
Fixes: #464

### DIFF
--- a/platform/tron/api.go
+++ b/platform/tron/api.go
@@ -280,7 +280,7 @@ func (p *Platform) GetBalance(address string) (string, error) {
 	}
 
 	for _, data := range account.Data {
-		return string(data.Balance), nil
+		return strconv.FormatUint(uint64(data.Balance), 10), nil
 	}
 	return "0", nil
 }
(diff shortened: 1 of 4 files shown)


platform/tron/api.go
```diff
@@ -274,11 +274,18 @@ func (p *Platform) GetDelegations(address string) (blockatlas.DelegationsPage, e
 }
 
 func (p *Platform) GetBalance(address string) (string, error) {
-	//TODO: Implement fetching balance for Tron
+	account, err := p.client.GetAccount(address)
+	if err != nil {
+		return "0", err
+	}
+
+	for _, data := range account.Data {
```
Pantani: if we have more them one `account.Data`?

Pantani:
do you need to fix unit tests, you can run using makefile: `make test`

platform/tron/api.go
dpereskokov: It is the same stuff as in Cosmos: https://github.com/trustwallet/blockatlas/blob/daec2de4faeac2e9815ec51d65a60aa12c408d6b/platform/cosmos/api.go#L133

I don't know why they enclose balances in the arrays perhaps for tokens transferring.

PR #468, opened 2019-10-16
Implement fetching balance for Tezos

dpereskokov:
Fixes: #463

### DIFF
--- a/platform/tezos/client.go
+++ b/platform/tezos/client.go
@@ -32,7 +32,7 @@ func (c *Client) GetBlockByNumber(num int64) ([]Tx, error) {
 
 func (c *Client) GetAccount(address string) (result Account, err error) {
 	path := fmt.Sprintf("node_account/%s", address)
-	referer := fmt.Sprint("https://tzscan.io/%s", address)
+	referer := fmt.Sprintf("https://tzscan.io/%s", address)
 	c.Headers["Referer"] = referer
 
 	return result, c.Get(&result, path, nil)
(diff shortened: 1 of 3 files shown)


platform/tezos/model.go
```diff
@@ -48,3 +48,17 @@ type Address struct {
...
+
+type DelegateOptions struct {
+	Setable bool `json:"setable"`
+}
+
+type Account struct {
+	Name          Address         `json:"name"`
+	Manager       Address         `json:"address"`
+	Delegate      DelegateOptions `json:"delegate"`
+	Balance       string          `json:"balance"`
```
vikmeup: can you only include what's needed? to keep models light

platform/tezos/client.go
```diff
@@ -49,3 +49,11 @@ func (c *Client) GetBlockByNumber(num int64) ([]Tx, error) {
 
 	return list, err
 }
+
+func (c *Client) GetAccount(address string) (result Account, err error) {
+	path := fmt.Sprintf("node_account/%s", address)
+	referer := fmt.Sprint("https://tzscan.io/%s", address)
```
vikmeup: why is this needed? the rest seems to work without this change

dpereskokov: This particular api returns 403 without this header. It was tricky part of that integration )

platform/tezos/model.go
dpereskokov: I believe we should describe models to allow us to do easy integrations in the future. Each time you want to know what this API returns you just have to pick in the sources and check. Other way you will be needed to have some pain finding api to trigger, fill it with referrer header to make them work etc.

vikmeup: in blockchain space most likely your code will break because of the changes you don't expect and to make it easier to maintain many platforms it's easier to keep only what's needed

PR #490, opened 2019-10-31
Try to get old address if multi address failed

hewigovens:


platform/ethereum/ens.go
```diff
@@ -34,6 +35,19 @@ func (p *Platform) Lookup(coin uint64, name string) (*blockatlas.Resolved, error
 
 	address, err := ensName.Address(coin)
 	if err != nil {
+		if coin == CoinType.ETH {
```
vikmeup: - can you move this into separate function? 
- can you add comment to remove this in the future?

PR #493, opened 2019-11-02
make middleware api reusable

Pantani:
move middleware and some gin stuff to pkg folder to be reusable to the market data application


Pantani:
> only naming changes?

No, I did create a `ginutils` package into the `pkg` folder and move the middlewares and the API's render stuffs to be reusable by another application.

PR #494, opened 2019-11-03
Add .xyz and .luxe for ens

hewigovens:


api/naming_service.go
```diff
@@ -29,6 +30,8 @@ func MakeLookupRoute(router gin.IRouter) {
 	ns.GET("/lookup", handleLookup)
 
 	TLDMapping[".eth"] = CoinType.ETH
+	TLDMapping[".xyz"] = CoinType.ETH
```
vikmeup: @hewigovens might need to drop `.` inside the mapping.

Does this mean anyone can create top level domains?

hewigovens: no, those two are supported by ens now, it can support more tlds

PR #495, opened 2019-11-04
show zero balance for Tezos delegated account

Pantani:


Pantani:
> Balance should be always as is
> 
> When you request delegation - that’s where you specify if it’s not available


@vikmeup  We only use this balance for the delegation. 
https://github.com/trustwallet/blockatlas/blob/master/api/batch.go#L46

We can change the method name for `UndelegatedBalance` or try to do something like `if coin == coin.Tezos().ID && len(delegations) > 0` to remove the balance. I prefer the first way.  Do you see another way?

vikmeup:
Renaming to `UndelegatedBalance` make sense, so we don't get confused in the future. 👍 

because we gonna use balance endpoint pretty soon for other use cases

PR #505, opened 2019-11-07
Add direction for non-UTXO's blockchain

Pantani:
# Headline
Add direction for non-UTXO's blockchains

## Problem
If you send a transaction for another wallet inside your app, you going two receive 2 push notifications with `received` title

https://github.com/trustwallet/backend/blob/master/services/notifications/message.go#L76

## Solution
Infer the direction for non-UTXO's blockchains, and parse inside the backend


Pantani:
> can we add unit test for this?

we need a big refactor for can test this method

PR #513, opened 2019-11-10
Add DigiByte (DGB)

kolya182:


coin/coins.go
```diff
@@ -501,6 +502,16 @@ var Coins = map[uint]Coin{
 		MinConfirmations: 0,
 		SampleAddr:       "nano_1trqphog5noig7z888asnjejcie8z1iopxyepcjdo1atps8whxiuwd51ehbw",
 	},
+	DGB: {
+		ID:               20,
+		Handle:           "digibyte",
+		Symbol:           "DGB",
+		Name:             "DigiByte",
+		Decimals:         8,
+		BlockTime:        15000,
```
hewigovens: what's the block time of DGB?

hewigovens:
Test failed

coin/coins.go
vikmeup: Sounds like 20 seconds or so

kolya182: 15 seconds https://www.digibyte.co/digibyte-global-blockchain

PR #516, opened 2019-11-13
Create storage for prices

Pantani:
- Create Redis storage methods for prices

### DIFF
--- a/pkg/blockatlas/marketdata.go
+++ b/pkg/blockatlas/marketdata.go
@@ -1,0 +1,58 @@
+package blockatlas
+
+import (
+	"math/big"
+	"time"
+)
+
+const (
+	TypeCoin  CoinType = "coin"
+	TypeToken CoinType = "token"
+)
+
+type CoinType string
+
+type TickerResponse struct {
+	Currency string  `json:"currency"`
+	Result   Tickers `json:"result"`
+}
+
+type Ticker struct {
+	Coin       uint        `json:"coin"`
+	CoinName   string      `json:"coin_name,omitempty"`
+	TokenId    string      `json:"token_id,omitempty"`
+	CoinType   CoinType    `json:"type,omitempty"`
+	Price      TickerPrice `json:"price,omitempty"`
+	LastUpdate time.Time   `json:"last_update,omitempty"`
+	Error      string      `json:"error,omitempty"`
+}
+
+type TickerPrice struct {
+	Value     *big.Float `json:"value,omitempty"`
+	Change24h *big.Float `json:"change_24h,omitempty"`
+	Currency  string     `json:"currency,omitempty"`
+	Provider  string     `json:"provider,omitempty"`
+}
+
+type Rate struct {
+	Currency  string  `json:"currency"`
+	Rate      float64 `json:"rate"`
+	Timestamp int64   `json:"timestamp"`
+}
+
+type Rates []Rate
+type Tickers []Ticker
+
+func (ts Tickers) ApplyRate(rate float64, currency string) {
+	for _, t := range ts {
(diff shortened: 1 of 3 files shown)


pkg/blockatlas/marketdata.go
```diff
@@ -0,0 +1,58 @@
+package blockatlas
+
```
vikmeup: why this needs to be exposed in blockatlas package? 

it might be better to have everything inside `pricing` module

pkg/blockatlas/marketdata.go
```diff
@@ -0,0 +1,58 @@
...
+type CoinType string
+
+type TickerResponse struct {
+	Currency string  `json:"currency"`
+	Result   Tickers `json:"result"`
+}
+
+type Ticker struct {
+	Coin       uint        `json:"coin"`
```
vikmeup: can we add `CoinType` as `uint` type so we can reuse across different models

pkg/blockatlas/marketdata.go
Pantani: because we going to use this model as API response

pkg/blockatlas/marketdata.go
```diff
@@ -0,0 +1,58 @@
...
+	CoinName   string      `json:"coin_name,omitempty"`
+	TokenId    string      `json:"token_id,omitempty"`
+	CoinType   CoinType    `json:"type,omitempty"`
+	Price      TickerPrice `json:"price,omitempty"`
+	LastUpdate time.Time   `json:"last_update,omitempty"`
+	Error      string      `json:"error,omitempty"`
+}
+
+type TickerPrice struct {
+	Value     *big.Float `json:"value,omitempty"`
```
vikmeup: any reason to use big float here? decimal values usually does not require big number, not for this use case

pkg/blockatlas/marketdata.go
Pantani:
```
type CoinType uint
const (
	TypeCoin  CoinType = 1
	TypeToken CoinType = 2
)
```

like that?

storage/market.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/trustwallet/blockatlas/pkg/logger"
+	"strings"
+)
+
+const (
+	EntityRates  = "currencies_rates"
+	EntityMarket = "market"
+)
+
+type ProviderList interface {
```
vikmeup: `PriceProvider`, `MarketProvider` might be more intuitive

pkg/blockatlas/marketdata.go
vikmeup: I was thinking to map somehow with this value: https://github.com/trustwallet/blockatlas/blob/master/coin/coins.go#L13

but not sure if that's possible

pkg/blockatlas/marketdata.go
Pantani: alway we need to use a number with a lot of decimals number and do some calc with them, it's better to use big numbers. It's more safety and you have less round and approximate values

pkg/blockatlas/marketdata.go
vikmeup: lgtm either way

vikmeup: Example from iOS:

Screen Shot 2019-11-12 at 5 10 50 PM

Pantani: we can do that, but we need to refactor the coin generator, it's easy, but a big change for this PR

vikmeup: I miss enums

Pantani: yes :( me too

PR #518, opened 2019-11-13
market/rates worker

Pantani:
- Create a worker logic to fetch prices and rates
- Create a generic object and interface for provider inherit
- Create a backoff method to try again exponentially
- The providers are missing


marketdata/rate/rate.go
```diff
@@ -0,0 +1,49 @@
...
+func (r *Rate) GetUpdateTime() time.Duration {
+	return r.UpdateTime
+}
+
+func (r *Rate) GetId() string {
+	return r.Id
+}
+
+func (r *Rate) GetType() string {
+	return "market-rate"
```
vikmeup: what types do you plan to have? besides pricing

Pantani: only for logs.... but I not happy with this as well

PR #522, opened 2019-11-13
fix api error response

Pantani:


pkg/ginutils/response.go
```diff
@@ -53,8 +53,14 @@ func (e *ApiError) Render() {
 	var msg map[string]interface{}
 	err := json.Unmarshal([]byte(e.StatusMessage), &msg)
 	if err == nil {
-		e.AbortWithStatusJSON(e.StatusCode, msg)
+		e.AbortWithStatusJSON(e.StatusCode, map[string]interface{}{
```
hewigovens: should we return 200 for error response?

Pantani: no, we going to return an error code, but the message going to be more friendly

PR #524, opened 2019-11-13
add fixer for price rates

Pantani:
Add currency rates provider from [Fixer](https://fixer.io/) service

### DIFF
--- a/marketdata/rate/fixer/fixer.go
+++ b/marketdata/rate/fixer/fixer.go
@@ -1,0 +1,46 @@
+package fixer
+
+import (
+	"github.com/spf13/viper"
+	"github.com/trustwallet/blockatlas/marketdata/rate"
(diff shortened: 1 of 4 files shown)


pkg/blockatlas/marketdata.go
```diff
@@ -35,24 +35,24 @@ type TickerPrice struct {
 }
 
 type Rate struct {
```
madcake: Where is it used?

Pantani: it's used by storage, market data, and API package 

Screen Shot 2019-11-13 at 14 34 31

madcake: Not for api result?

Pantani: no, but we can create a route to retrieve the rate as well, now we are using inside the API package only to apply the currency rate in API result

PR #525, opened 2019-11-13
remove tezos page limit

Pantani:
remove tezos page limit for transactions

### DIFF
--- a/platform/tezos/api.go
+++ b/platform/tezos/api.go
@@ -91,7 +91,7 @@ func NormalizeDelegation(account Account, validators blockatlas.ValidatorMap) ([
 func NormalizeTxs(srcTxs []Tx) (txs []blockatlas.Tx) {
 	for _, srcTx := range srcTxs {
 		tx, ok := NormalizeTx(&srcTx)
-		if !ok || len(txs) >= blockatlas.TxPerPage {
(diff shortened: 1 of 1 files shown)


platform/tezos/client.go
```diff
@@ -14,7 +14,7 @@ type Client struct {
 func (c *Client) GetTxsOfAddress(address string) ([]Tx, error) {
 	var account Op
 	path := fmt.Sprintf("account/%s/op", address)
-	err := c.Get(&account, path, url.Values{"limit": {"1000"}, "offset": {"0"}})
+	err := c.Get(&account, path, url.Values{"limit": {"5000"}, "offset": {"0"}})
```
madcake: Why?

kolya182: Yehh, why not just query as much as possible and limit to TxPage size ?

Pantani: For Tezos, some operations are not parsed because is not a transaction, we can miss, but I going to revert, @ArtemGoryunov made this same question

Pantani: Make sense we use 5k, but the Tezos infra it's not good now, we can have some timeouts

madcake: How much returned by default?

PR #532, opened 2019-11-15
small fixes for prices sync

Pantani:
- Fix priority filter
- Fix CMC parser
- Run schedule in the background
- Add env var to update time

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -34,14 +34,14 @@ market:
   enabled: false
   auth: test
   dex_api: https://dex.binance.org/api
-  dex_quote_update_seconds: 240
+  dex_quote_update_time: 2m
   cmc_api: https://pro-api.coinmarketcap.com
   cmc_api_key:
-  cmc_quote_update_seconds: 360
-  cmc_rate_update_seconds: 360
+  cmc_quote_update_time: 5m
+  cmc_rate_update_time: 5m
   fixer_api: https://data.fixer.io/api
   fixer_key:
-  fixes_rate_update_seconds: 240
+  fixer_rate_update_time: 1h
 
 storage:
   redis: redis://localhost:6379
--- a/config.yml
+++ b/config.yml
@@ -34,14 +34,14 @@ market:
   enabled: false
   auth: test
   dex_api: https://dex.binance.org/api
-  dex_quote_update_seconds: 30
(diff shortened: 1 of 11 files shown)


config.yml
```diff
@@ -34,10 +34,14 @@ market:
   enabled: false
   auth: test
   dex_api: https://dex.binance.org/api
+  dex_quote_update_seconds: 30
   cmc_api: https://pro-api.coinmarketcap.com
   cmc_api_key:
+  cmc_quote_update_seconds: 30
```
vikmeup: cmc should be 5 minutes at least

config.yml
```diff
@@ -34,10 +34,14 @@ market:
...
+  cmc_rate_update_seconds: 30
   fixer_api: https://data.fixer.io/api
   fixer_key:
+  fixes_rate_update_seconds: 30
```
vikmeup:
```suggestion
  fixer_rate_update_seconds: 30
```

also it should be ~1 hour? No need to update that often as rates don't change much

marketdata/market/dex/dex.go
```diff
@@ -20,13 +20,14 @@ type Market struct {
 }
 
 func InitMarket() market.Provider {
+	update := viper.GetDuration("market.dex_quote_update_seconds")
 	m := &Market{
 		Market: market.Market{
 			Id:         "dex",
 			Name:       "Binance Dex",
 			URL:        "https://www.binance.org/",
```
vikmeup: why do we pass market url here?

marketdata/market/dex/dex.go
```diff
@@ -20,13 +20,14 @@ type Market struct {
 }
 
 func InitMarket() market.Provider {
+	update := viper.GetDuration("market.dex_quote_update_seconds")
 	m := &Market{
 		Market: market.Market{
 			Id:         "dex",
 			Name:       "Binance Dex",
```
vikmeup: is name required param?

Pantani: no, name and url it's optional. Id and API it's required

marketdata/market/dex/dex.go
Pantani: I did an API to return all provider, but it's useless and I did remove

Issue #85, opened 2019-04-25
Implement NEW_ORDER, CANCEL_ORDER transaction type


kolya182:
https://explorer.binance.org/api/v1/txs?address=bnb16ya67j7kvw8682kka09qujlw5u7lf4geqef0ku&page=1&rows=100

https://developer.trustwallet.com/blockatlas/transaction-format#any-action
```
{
    "type": "any_action",
    "metadata": {
       "title": "Place Order",
       "key":  "place_order",     
    }
}
```

kolya182:
Part of https://github.com/trustwallet/blockatlas/issues/123

PR #545, opened 2019-11-18
Parse BNB order transactions

Pantani:
- Refactor `NormalizeTx`;
- Refactor `api_test.go`;
- Parse `NEW_ORDER` and `CANCEL_ORDER` transactions;

closes https://github.com/trustwallet/blockatlas/issues/85


pkg/blockatlas/tx.go
```diff
@@ -149,14 +147,14 @@ type ContractCall struct {
...
-	TokenID  string `json:"token_id"`
-	Name     string `json:"name"`
-	Symbol   string `json:"symbol"`
-	Decimals uint   `json:"decimals"`
-	Value    Amount `json:"value"`
+	Coin     uint        `json:"coin"`
+	Title    string      `json:"title"`
+	Key      KeyType     `json:"key"`
+	TokenID  string      `json:"token_id,omitempty"`
+	Name     string      `json:"name,omitempty"`
```
vikmeup: iOS requires those values, otherwise it won't be able to parse it

Screen Shot 2019-11-17 at 5 59 55 PM

pkg/blockatlas/tx.go
```diff
@@ -64,7 +62,7 @@ type Tx struct {
 	// Address of the transaction sender
 	From string `json:"from"`
 	// Address of the transaction recipient
-	To string `json:"to"`
+	To string `json:"to,omitempty"`
```
vikmeup: to is required param as well on clients side

pkg/blockatlas/tx.go
```diff
@@ -87,7 +85,7 @@ type Tx struct {
 	// Transaction Direction
 	Direction Direction `json:"direction,omitempty"`
 	// Meta data object
-	Memo string      `json:"memo"`
+	Memo string      `json:"memo,omitempty"`
```
vikmeup: memo is required on client side

platform/binance/model.go
```diff
@@ -53,6 +63,45 @@ type Tx struct {
 	Memo          string      `json:"memo"`
 }
 
+func (tx *Tx) getData() (*Data, error) {
```
dpereskokov: do we need pointer here?

dpereskokov: let write tests for this function?

platform/binance/model_test.go
```diff
@@ -0,0 +1,55 @@
...
+func TestTx_getData(t *testing.T) {
+	tests := []struct {
+		name    string
+		Data    string
+		want    Data
+		wantErr bool
+	}{
+		{
+			"new order",
+			"{\"orderData\":{\"symbol\":\"AWC-986_BNB\",\"orderType\":\"limit\",\"side\":\"buy\",\"price\":0.00324939,\"quantity\":2.00000000,\"timeInForce\":\"GTE\",\"orderId\":\"D13BAF4BD6638FA3AAD6EBCA0E4BEEA73DF4D519-30\"}}",
```
vikmeup: @Pantani should we suggest binance dex folks to have this as an object and not as a string?

Pantani: definitely yes 👍

PR #546, opened 2019-11-18
use float instead string for market response

Pantani:
# Headline
Use float instead of string for market response

# Problem
We cannot serialize a `big.Float`  into a number, only into a string

# Solution
Change `big.Float` to `float64`

### DIFF
--- a/cmd/sync.go
+++ b/cmd/sync.go
@@ -2,7 +2,9 @@ package cmd
 
 import (
 	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
 	"github.com/trustwallet/blockatlas/marketdata"
+	"github.com/trustwallet/blockatlas/pkg/logger"
 )
 
 var syncCmd = &cobra.Command{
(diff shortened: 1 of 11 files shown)


cmd/sync.go
```diff
@@ -13,6 +15,10 @@ var syncCmd = &cobra.Command{
 }
 
 func syncMarketData(cmd *cobra.Command, args []string) {
+	if !viper.GetBool("market.enabled") {
```
vikmeup: I think we should check this in higher function, not inside

marketdata/market/dex/dex.go
```diff
@@ -41,7 +40,7 @@ func (m *Market) GetData() (blockatlas.Tickers, error) {
 		return nil, errors.E(err, "rate not found", errors.Params{"asset": BNBAsset})
 	}
 	result := normalizeTickers(prices, m.GetId())
-	result.ApplyRate(rate.Rate.Quo(big.NewFloat(1), rate.Rate), "USD")
+	result.ApplyRate(1/rate.Rate, "USD")
```
vikmeup: Can you add USD to const?

cmd/sync.go
Pantani: inside `root.go`? How do we check this?

PR #552, opened 2019-11-19
Implement batch fetching ens addresses

dpereskokov:
Fixes: #551


api/batch.go
```diff
@@ -81,3 +88,51 @@ func makeCategoriesBatchRoute(router gin.IRouter) {
...
+// @Router /v2/ns/lookup [post]
+func makeNsLookupRoute(router gin.IRouter) {
+	router.POST("/ns/lookup", func(c *gin.Context) {
+		var req ENSBatchRequest
+		if err := c.BindJSON(&req); err != nil {
+			ginutils.ErrorResponse(c).Message(err.Error()).Render()
+			return
+		}
+
+		lookupsChan := make(chan blockatlas.Resolved)
```
Pantani: @dpereskokov can you open other PR to make the others batch request asynchronous? 😄

api/naming_service.go
```diff
@@ -27,40 +30,45 @@ var TLDMapping = map[string]uint64{}
...
 	TLDMapping[".zil"] = CoinType.ZIL
 }
 
-func handleLookup(c *gin.Context) {
-	name := c.Query("name")
-	coinQuery := c.Query("coin")
-
+func handleLookup(name string, coin uint64) (blockatlas.Resolved, error) {
+	result := blockatlas.Resolved{Result: "", Coin: coin}
 	if name == "" {
```
Pantani: can we move this verification to the top-level function? I think it's not the responsibility of this method to verify this. 

Just a suggestion

api/naming_service.go
```diff
@@ -27,40 +30,45 @@ var TLDMapping = map[string]uint64{}
...
 
+	name = strings.ToLower(name)
 	for tld, id := range TLDMapping {
 		if strings.HasSuffix(name, tld) {
 			api := platform.NamingAPIs[id]
 			resolved, err := api.Lookup(coin, name)
 			if err != nil {
-				ginutils.RenderError(c, http.StatusBadRequest, err.Error())
+				return result, err
 			} else {
```
Pantani: you don't need this else

api/batch.go
dpereskokov: This one is really time consuming. Checking one coin takes ~5 seconds

api/naming_service.go
dpereskokov: Hm, do you think so? top level function just parses query data in this case. And this func is pure and checks if parameters passed to it are normal. It's also used for batch request.

api/batch.go
```diff
@@ -81,3 +88,51 @@ func makeCategoriesBatchRoute(router gin.IRouter) {
...
+		var wg sync.WaitGroup
+		wg.Add(len(req.Coins))
+
+		for _, coin := range req.Coins {
+			go func(coin uint64) {
+				defer wg.Done()
+				lookup, err := handleLookup(req.Name, coin)
+				if err != nil {
+					logger.Error(err)
```
Pantani:
```suggestion
					logger.Error(err, logger.Params{"name": req.Name, "coin": coin, })
```

api/naming_service.go
Pantani: yes, The top level is for inputs, we get all we need to parse and pass to the method to do only the method need to do

api/batch.go
Pantani: put more context inside the logs

api/naming_service.go
dpereskokov: The function can be viewed as an individual component. Its author is best placed to define any preconditions and check them. Checking them outside presupposes the caller knows the preconditions which may not be the case. Also by placing them inside the function you're assured every call is checked. In function body usually perfomed checks on any post-conditions before leaving the function.

dpereskokov: Just the case: imagine you have function calcSquare(x, y). Who is responsible to check x and y are not negative?

api/batch.go
hewigovens: I don't this is the best way to handle batch query, instead of calling `handleLookup` one by one, we should add `handleBatchLookup` (or something smilar) to `ens.go` and `zns.go`

For example, we can try batch JSON RPC request for ENS, and for UD, if you take a look at the response, we don't need to send multiple request to get all the coins

api/naming_service.go
madcake: Bullshit man. Why you do something and on next step throw exception? If your function as individual component you shouldt check data in function. In design theory you should thing that your function live in perfect place.
Looked other languages: swift kotlin, java. You can say that function should get only non null values.

@Pantani have good suggestion. It's error check name in this function.

dpereskokov: @madcake Good design is to check parameters in every place before passing them? In our case we have to do it twice (for single coin handler and for batch). 
Do you think we have to delegate checking functions parameters logic for somebody's else? Function that handles api call have no idea what is going to be done in the `handleLookup` but anyway has to somehow validate its parameters instead of just to be transparent and pass them through? What if it will be possible to pass only names prefixed with 'joe'. Do you think we have to handle this condition in every place somewhere but not in our func?

Why standard GO libraries return errors then? For example json.Unmarshal. Based on what you've just said we have to check that we pass valid json there by ourselves every time.

Pantani: @dpereskokov The function doesn't need to verify the params, this is not a good pattern in every language, we need to verify the inputs at a high level. When we receive these inputs, the functions need to do only the thing to in charge, not to verify the parameters. Verification needs to be done in the parse

Pantani: > Just the case: imagine you have function calcSquare(x, y). Who is responsible to check x and y are not negative?

If the software is only to do that, make sense, but it is only a method in the code, and do you need to verify always do call this method, you gonna do a lot of duplicated code

The code needs to return an error and we handle the error. This is the reason the Go is oriented to errors. `calcSquare(x, y) (float64, error)`

Pantani: param validation inside the method needs to be an exception. The input always need to be verified at the top level, and the low levels need to do only the things they born to do

pkg/blockatlas/api.go
```diff
@@ -70,5 +70,5 @@ type CustomAPI interface {
 
 type NamingServiceAPI interface {
 	Platform
-	Lookup(coins []uint64, name string) ([]Resolved, error)
+	Lookup(coins []int, name string) ([]Resolved, error)
```
vikmeup: can we use `uint64 ` ?

PR #553, opened 2019-11-19
improve tx sort

Pantani:
improve tx sort and remove an unused package

### DIFF
--- a/pkg/blockatlas/marshal.go
+++ b/pkg/blockatlas/marshal.go
@@ -3,6 +3,7 @@ package blockatlas
 import (
 	"encoding/json"
 	"github.com/trustwallet/blockatlas/pkg/errors"
+	"github.com/trustwallet/blockatlas/pkg/logger"
 	"github.com/trustwallet/blockatlas/util"
 	"regexp"
 	"sort"
@@ -110,7 +111,9 @@ func (a *Amount) MarshalJSON() ([]byte, error) {
 
 // Sort sorts the response by date, descending
 func (txs *TxPage) Sort() {
+	logger.Info(txs)
 	sort.Sort(txs)
+	logger.Info(txs)
 }
 
 // MarshalJSON returns a wrapped list of transactions in JSON
--- a/pkg/blockatlas/marshal.go
+++ b/pkg/blockatlas/marshal.go
@@ -2,7 +2,6 @@ package blockatlas
 
 import (
 	"encoding/json"
-	"github.com/spf13/cast"
 	"github.com/trustwallet/blockatlas/pkg/errors"
 	"github.com/trustwallet/blockatlas/util"
 	"regexp"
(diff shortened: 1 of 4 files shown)


pkg/blockatlas/marshal.go
```diff
@@ -110,12 +109,8 @@ func (a *Amount) MarshalJSON() ([]byte, error) {
...
 
 // Sort sorts the response by date, descending
-func (r *TxPage) Sort() {
-	sort.Slice(*r, func(i, j int) bool {
-		ti := cast.ToUint64((*r)[i].Date)
-		tj := cast.ToUint64((*r)[j].Date)
-		return ti >= tj
-	})
+func (txs *TxPage) Sort() {
+	sort.Sort(txs)
```
vikmeup: do we have test for this?  good either way

pkg/blockatlas/tx.go
```diff
@@ -53,6 +53,10 @@ const TxPerPage = 25
 // TxPage is a page of transactions
 type TxPage []Tx
 
+func (txs TxPage) Len() int           { return len(txs) }
+func (txs TxPage) Swap(i, j int)      { txs[i], txs[j] = txs[j], txs[i] }
```
vikmeup: is this some type of protocol to support `Sort` ?

Pantani: yes!!!

pkg/blockatlas/marshal.go
Pantani: I can do that

pkg/blockatlas/tx.go
vikmeup: Good to know, thanks

pkg/blockatlas/marshal.go
```diff
@@ -110,7 +111,9 @@ func (a *Amount) MarshalJSON() ([]byte, error) {
 
 // Sort sorts the response by date, descending
 func (txs *TxPage) Sort() {
+	logger.Info(txs)
```
vikmeup: remove logger?

pkg/blockatlas/marshal_test.go
```diff
@@ -75,3 +75,25 @@ func TestTx_MarshalJSON(t *testing.T) {
...
+func TestTxPage_Sort(t *testing.T) {
+	tests := []struct {
+		name string
+		txs  TxPage
+		want TxPage
+	}{
+		{
+			"sort 1",
+			TxPage{{Date: 9999}, {Date: 46}, {Date: 100}, {Date: 1}, {Date: 777}},
+			TxPage{{Date: 1}, {Date: 46}, {Date: 100}, {Date: 777}, {Date: 9999}},
```
vikmeup: I think we need transactions to be sorted in the opposite way? 

Since it's a unix timestramp we need them to be sorted from old to new?

PR #554, opened 2019-11-19
fix bnb fee

Pantani:


Pantani:
> Maybe we should add handling in `DecimalExp` itself?
> 
> Unit test for this would be useful

we already have, but it doesn't complain about the `.` after the zero

vikmeup:
Where is the `0.` come from?

Pantani:
> Where is the `0.` come from?

when the `srcTx.Fee` is zero, the `DecimalExp` format the fee to `0.`

Pantani:
I'm going merge because we need that, we can improve later

vikmeup:
Can we add test case for `DecimalExp` to handle this? it sounds like a bug inside `DecimalExp` itself

Issue #473, opened 2019-10-20
Replace end off life Veforge explorer


kolya182:
https://www.reddit.com/r/CryptoCurrency/comments/dkigf4/veforge_announced_the_shutdown_of_their_block/?utm_source=share&utm_medium=ios_app&utm_name=iossmf

https://medium.com/@vechainofficial/announcing-the-new-official-vechain-block-explorer-and-vechain-sync-ledger-wallet-interface-38a4b5360aeb

vikmeup:
@kolya182 what's the status on this?

kolya182:
Scheduled for this week

Pantani:
https://github.com/trustwallet/blockatlas/pull/557

PR #557, opened 2019-11-20
[WIP] new vechain platform

Pantani:
Remove Veforge and use RPC API's for Vechain

closes https://github.com/trustwallet/blockatlas/issues/473


config.yml
```diff
@@ -125,7 +125,7 @@ tron:
 
 # [VET] VeChain: https://www.vechain.org
 vechain:
- api: https://explore.veforge.com/api
+  api: https://vethor-pubnode.digonchain.com
```
kolya182: Why do we call it `api` if in reality we call `rpc` ?. I understand meaning for explorers as 3rd party solution.

util/string.go
```diff
@@ -0,0 +1,8 @@
+package util
+
+func GetValidParameter(first, second string) string {
```
kolya182: What does it do ?

config.yml
Pantani: this is an RPC node API

util/string.go
Pantani:
```
	if len(first) > 0 {
		return first
	}
	return second
```

Pantani: I did remove from Ethereum platform because I need reuse that

https://github.com/trustwallet/blockatlas/pull/557/files#diff-ce80ab28ecba3e974a0b3b36b8c16297L8

platform/vechain/api_test.go
```diff
@@ -88,15 +88,15 @@ const transferLogSrc = `{
...
-	Type:     "transfer",
-	Fee:      blockatlas.Amount("0"),
-	Status:   blockatlas.StatusCompleted,
-	Block:    701968,
+	ID:     "0xc89b224842f3b5edcf4e6950194148be7e83cc951ff4fdfa2490e4c0cf12c80b",
+	Coin:   coin.VET,
+	From:   "0x15f4d9bed894e2e426d65e3df1480c61fb131a57",
+	To:     "0x15f4d9bed894e2e426d65e3df1480c61fb131a57",
+	Date:   1537342090,
+	Type:   "transfer",
```
vikmeup: can we use type for transfer?

util/string.go
Pantani: https://github.com/trustwallet/blockatlas/pull/557/files#diff-f240f9e265c26fdb22451440b38e2f58R14

Issue #209, opened 2019-07-28
[Observer] Add TRON support for BlockAPI


vikmeup:
curl -X POST  https://api.trongrid.io/wallet/getblockbynum -d '{"num":10874448}'

kolya182:
From Tronscan api https://github.com/tronscan/tronscan-frontend/wiki/TRONSCAN-API#22

Get block info https://apilist.tronscan.org/api/block?number=5987471 use `timestamp to query transactions` https://apilist.tronscan.org/api/transaction?sort=-timestamp&count=true&limit=20&start=0&start_timestamp=1548056253000&end_timestamp=1548056253000 => 64 == 64 (https://tronscan.org/#/block/5987471)

PR #559, opened 2019-11-21
Add Tron block API

Pantani:
- Add block API support for TRON
- Refactor TRON API
- Split stake stuffs in a separate file

closes https://github.com/trustwallet/blockatlas/issues/209


vikmeup:
Does this support only TRX transfers but also TRC10?

Pantani:
> Does this support only TRX transfers but also TRC10?


Only TRX, I going see how to filter the TRC10 txs

platform/tron/api.go
```diff
@@ -27,6 +27,32 @@ func (p *Platform) Coin() coin.Coin {
 	return coin.Coins[coin.TRX]
 }
 
+func (p *Platform) CurrentBlockNumber() (int64, error) {
+	return p.client.CurrentBlockNumber()
+}
+
+func (p *Platform) GetBlockByNumber(num int64) (*blockatlas.Block, error) {
```
dpereskokov: do we need pointer here?

platform/tron/api.go
```diff
@@ -82,10 +82,9 @@ func (p *Platform) GetTokenTxsByAddress(address, token string) (blockatlas.TxPag
 
 	var tokenInfo AssetInfo
 	info, err := p.client.GetTokenInfo(token)
-	if err != nil && len(info.Data) == 0 {
+	if err != nil || len(info.Data) == 0 {
```
vikmeup: is this intended? it might cause issues

platform/tron/stake.go
```diff
@@ -0,0 +1,102 @@
+package tron
+
+import (
+	"github.com/trustwallet/blockatlas/pkg/blockatlas"
+	"github.com/trustwallet/blockatlas/pkg/logger"
+	services "github.com/trustwallet/blockatlas/services/assets"
```
vikmeup: I like this idea to keep in separate file as `tron/stake.go` for each interface, makes it clean

Pantani: yes, I did start the refactor of the API, but I'm starting to be crazy with a lot of methods the `api.go` has. I did split the stake things to be more clear, we can do that for every API

platform/tron/api.go
Pantani: yes! to satisfy the platform methods inside the registry. And this pointer going to stay at a high level, don't have a chance to the garbage collector can't desalloc that

platform/tron/api.go
Pantani: yes, the old logic doesn't make a lot of sense

Pantani:
> Does this support only TRX transfers but also TRC10?

@vikmeup now support!!!!

Issue #561, opened 2019-11-21
Missing "Place Order" type transactions on BEP2 token screen


ArtemGoryunov:
## Summary of Bug

https://blockatlas.trustwalletapp.com/v1/binance/bnb1xwalxpaes9r0z0fqdy70j3kz6aayetegur38gl?token=AWC-986

returns

`{"total":0,"docs":[],"status":true}`
that's why it is empy.

![69286103-38e1fd80-0ba7-11ea-887d-c15dbd848cab](https://user-images.githubusercontent.com/725247/69337003-e10fc900-0c70-11ea-9a75-56582b100110.jpeg)

## Steps to Reproduce

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #562, opened 2019-11-22
fix bnb token history

Pantani:
closes https://github.com/trustwallet/blockatlas/issues/561


Pantani:
> But this adds orders to the list? which we should keep disabled until update clients
> 
> test would be useful here

No, only if you filter the list by the token. We already have tests here, but don't cover this case, I going to increase that

PR #565, opened 2019-11-24
Create a cache for the validator map

Pantani:
# Headline
- Create a GET and POST methods to do requests with cache
- Create a cache for the validator's map

### DIFF
--- a/pkg/blockatlas/client.go
+++ b/pkg/blockatlas/client.go
@@ -19,6 +19,7 @@ type Request struct {
 	Headers      map[string]string
 	HttpClient   *http.Client
 	ErrorHandler func(res *http.Response, uri string) error
+	CacheHandler func(method string, url string, body io.Reader, result interface{}) error
 }
 
 func InitClient(baseUrl string) Request {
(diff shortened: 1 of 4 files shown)


services/assets/client.go
```diff
@@ -79,3 +94,23 @@ func NormalizeValidator(plainValidator blockatlas.Validator, validator AssetVali
 func GetImage(c coin.Coin, ID string) string {
 	return AssetsURL + c.Handle + "/validators/assets/" + ID + "/logo.png"
 }
+
+func cacheValidatorList(method string, url string, body io.Reader, result interface{}) error {
```
vikmeup: is this possible to make as generic method? that could be used in other places

pkg/blockatlas/client.go
```diff
@@ -15,10 +15,12 @@ import (
...
 type Request struct {
-	BaseUrl      string
-	Headers      map[string]string
-	HttpClient   *http.Client
-	ErrorHandler func(res *http.Response, uri string) error
+	BaseUrl            string
+	Headers            map[string]string
+	HttpClient         *http.Client
+	ErrorHandler       func(res *http.Response, uri string) error
+	CacheHandler       func(method string, url string, body io.Reader, result interface{}) error
```
vikmeup: is there a reason to have `CacheHandler`? I would assume most requests would just need to specify expiration time and not need for custom handlers or when handler might different?

PR #566, opened 2019-11-25
fix dex quote and the API response parameter 

Pantani:
fix dex quote and the API response parameter for markets

### DIFF
--- a/marketdata/market/dex/dex.go
+++ b/marketdata/market/dex/dex.go
@@ -59,9 +59,9 @@ func normalizeTicker(price CoinPrice, provider string) (blockatlas.Ticker, error
 		return blockatlas.Ticker{}, errors.E(err, "normalizeTicker parse value24h error",
 			errors.Params{"PriceChange": price.PriceChangePercent, "Symbol": price.BaseAssetName})
 	}
-	tokenId := price.BaseAssetName
+	tokenId := price.QuoteAssetName
 	if tokenId == BNBAsset {
-		tokenId = price.QuoteAssetName
+		tokenId = price.BaseAssetName
 		value = 1.0 / value
 	}
 	return blockatlas.Ticker{
--- a/pkg/blockatlas/marketdata.go
+++ b/pkg/blockatlas/marketdata.go
@@ -15,7 +15,7 @@ type CoinType string
 
 type TickerResponse struct {
 	Currency string  `json:"currency"`
-	Result   Tickers `json:"result"`
+	Prices   Tickers `json:"prices"`
 }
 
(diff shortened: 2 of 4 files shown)


pkg/blockatlas/marketdata.go
```diff
@@ -15,7 +15,7 @@ type CoinType string
 
 type TickerResponse struct {
 	Currency string  `json:"currency"`
-	Result   Tickers `json:"result"`
+	Prices   Tickers `json:"prices"`
```
vikmeup: should this be `docs` ? for consistency

marketdata/market/dex/dex.go
```diff
@@ -59,9 +59,9 @@ func normalizeTicker(price CoinPrice, provider string) (blockatlas.Ticker, error
 		return blockatlas.Ticker{}, errors.E(err, "normalizeTicker parse value24h error",
 			errors.Params{"PriceChange": price.PriceChangePercent, "Symbol": price.BaseAssetName})
 	}
-	tokenId := price.BaseAssetName
+	tokenId := price.QuoteAssetName
 	if tokenId == BNBAsset {
-		tokenId = price.QuoteAssetName
+		tokenId = price.BaseAssetName
```
vikmeup: Is there any issues if one token could be pair to others? 

Like XXX could be listed on markets:
XXX-BNB
XXX-USDC

pkg/blockatlas/marketdata.go
Pantani: @lorenzozanotto ask me to change that to prices

marketdata/market/dex/dex.go
Pantani: No, this condition solve this problem, but the Dex prices are wrong because the logic is inverted

pkg/blockatlas/marketdata.go
vikmeup: I think he meant `docs`

we should either keep `result` or `docs`

lorenzozanotto: @vikmeup @Pantani I asked to use `prices` because I took that from the `PaginationResponsePrices` struct used to draw the price charts from an array of prices. 

Using `docs` would be more consistent, I agree. Should we change also the field on the `/charts"` endpoint response in the TrustAPI as well? That's the only one call which uses `prices` instead of `docs` for a paginated response

vikmeup: `/charts` will be re-build to blockatlas within the next 2-3 weeks, so we can adopt `docs` model as well.

PR #568, opened 2019-11-25
Priority to fetch rates

Pantani:
Create a priority to fetch rates for fiat currency inside the market data API, like to fetch the quotes

### DIFF
--- a/marketdata/rate/coinmarketcap/cmc_test.go
+++ b/marketdata/rate/coinmarketcap/cmc_test.go
@@ -10,6 +10,7 @@ import (
 )
 
 func Test_normalizeRates(t *testing.T) {
+	provider := "cmc"
 	tests := []struct {
 		name      string
 		prices    CoinPrices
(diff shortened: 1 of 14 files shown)


marketdata/rate/coinmarketcap/cmc_test.go
```diff
@@ -10,6 +10,7 @@ import (
 )
 
 func Test_normalizeRates(t *testing.T) {
+	provider := "cmc"
```
vikmeup: Is this string somewhere else? Maybe add as const?

pkg/blockatlas/marketdata.go
```diff
@@ -46,6 +46,7 @@ type Rate struct {
 	Currency  string  `json:"currency"`
 	Rate      float64 `json:"rate"`
 	Timestamp int64   `json:"timestamp"`
+	Provider  string  `json:"provider,omitempty"`
```
vikmeup: Should we add name Provider?

storage/market.go
```diff
@@ -43,15 +43,28 @@ func (s *Storage) GetTicker(coin, token string) (blockatlas.Ticker, error) {
...
+type RateProviderList interface {
+	GetPriority(providerId string) int
+}
+
+func (s *Storage) SaveRates(rates blockatlas.Rates, pl RateProviderList) {
 	for _, rate := range rates {
 		r, err := s.GetRate(rate.Currency)
-		if err == nil && rate.Timestamp < r.Timestamp {
-			return
+		if err == nil {
```
dpereskokov: Does it make sense to move this condition to separate func since it's been used twice?

marketdata/rate/coinmarketcap/cmc_test.go
Pantani: we only use this string here

pkg/blockatlas/marketdata.go
Pantani: this parameter doesn't be returned by the API, only to save the rates in the database and keep the highest priority

storage/market.go
Pantani: the objects are different, I tried to do that, but we need to create some interfaces and the code starts to be bigger

PR #585, opened 2019-11-27
add tx direction for transaction API

Pantani:


Pantani:
> Do we have similar logic for push notifications?
> 
> Could this logic be reused? And how UTXO based chains are handled?
> 
> Unit test might be great here

Yes we have similar logic for push notifications, I going to split that and create some unit tests

The utxo blockchain is handled by another route
`/v1/bitcoin/xpub/zpub6rYtS3t7AbXkgZnPrwo433VuSNLGfLRPP192Enh7Pn9Dw82NuAQjXXQY8TtdJKf1g1ZNRdhBvaGAbgcAsaPAUpkNR8kMZ55vCi8thxHQrXo`
`/v1/bitcoin/address/bc1qw8a5lcpnt32kshmvxnxakcc9c4mapdvflxcnw4`

PR #586, opened 2019-11-27
Add token_id for VIP180 tokens

kolya182:


platform/vechain/api_test.go
```diff
@@ -140,7 +140,7 @@ var expectedTransferLog = blockatlas.TxPage{
 		Meta: blockatlas.TokenTransfer{
 			Name:     "",
 			Symbol:   "VTHO",
-			TokenID:  "",
+			TokenID:  "0x0000000000000000000000000000456e65726779",
```
vcoolish: this is lowercase token id

vikmeup: It should be `0x0000000000000000000000000000456E65726779`

Issue #598, opened 2019-11-30
Add pricing support for ctoken / compound


vikmeup:
https://api.compound.finance/api/v2/ctoken

https://compound.finance/developers/api#CTokenService

PR #606, opened 2019-12-04
Add pricing support for ctoken / compound

dpereskokov:
Fixes: #598

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -32,22 +32,18 @@ observer:
 # The market watcher
 market:
   enabled: false
+  quote_update_time: 20m
+  rate_update_time: 20m
   dex:
-    quote_update_time: 20m
     api: https://dex.binance.org/api
   cmc:
-    rate_update_time: 15m
-    quote_update_time: 15m
     api: https://pro-api.coinmarketcap.com
     api_key:
     map_url: https://raw.githubusercontent.com/trustwallet/assets/master/pricing/coinmarketcap
   fixer:
-    rate_update_time: 1h
     api: https://data.fixer.io/api
     api_key:
   compound:
-    rate_update_time: 15m
-    quote_update_time: 15m
     api: https://api.compound.finance/api
 
 storage:
--- a/config.yml
+++ b/config.yml
@@ -45,6 +45,10 @@ market:
     rate_update_time: 1h
     api: https://data.fixer.io/api
     api_key:
+  compound:
+    rate_update_time: 15m
+    quote_update_time: 15m
+    api: https://api.compound.finance/api
 
 storage:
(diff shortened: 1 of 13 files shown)


marketdata/market/compound/compound.go
```diff
@@ -0,0 +1,63 @@
...
+
+import (
+	"github.com/spf13/viper"
+	"github.com/trustwallet/blockatlas/marketdata/market"
+	"github.com/trustwallet/blockatlas/pkg/blockatlas"
+	"time"
+)
+
+const (
+	ETHAsset = "ETH"
```
vikmeup: use `coin` package to get symbol of `ETH`

config.yml
```diff
@@ -45,6 +45,10 @@ market:
     rate_update_time: 1h
     api: https://data.fixer.io/api
     api_key:
+  compound:
+    rate_update_time: 15m
+    quote_update_time: 15m
```
vikmeup: can we use generic timeout? 

`market.rate_update_time `
`quote.quote_update_time `

this might be reused quite a lot by other providers

vikmeup: Let's always use standard timeout, unless there is a need for custom

marketdata/market/compound/compound.go
```diff
@@ -0,0 +1,63 @@
...
+)
+
+type Market struct {
+	market.Market
+}
+
+func InitMarket() market.Provider {
+	m := &Market{
+		Market: market.Market{
+			Id:         "compound",
```
vikmeup: can you add `"compound"` to `const`?

marketdata/market/compound/compound.go
vikmeup: `coin.Ethereum().Symbol`

config.yml
```diff
@@ -32,22 +32,18 @@ observer:
...
-    quote_update_time: 20m
     api: https://dex.binance.org/api
   cmc:
-    rate_update_time: 15m
-    quote_update_time: 15m
     api: https://pro-api.coinmarketcap.com
     api_key:
     map_url: https://raw.githubusercontent.com/trustwallet/assets/master/pricing/coinmarketcap
   fixer:
-    rate_update_time: 1h
```
vikmeup: we should keep fixer as `1h`, as it does not require fast updates

config.yml
```diff
@@ -32,22 +32,18 @@ observer:
 # The market watcher
 market:
   enabled: false
+  quote_update_time: 20m
+  rate_update_time: 20m
   dex:
-    quote_update_time: 20m
```
vikmeup: we should set dex price update to 5m

PR #609, opened 2019-12-05
Fix nano transactions history

kolya182:
Root issues fixed in this commit https://github.com/trustwallet/blockatlas/commit/e8b70a8616d1257be32aaad61d0d43eb5a16d0ec
I believe @hewigovens implemented logic to work with rpc 
https://github.com/trustwallet/blockatlas/commit/481c807babbac01a9087a3f3bc728771d637e2ee#diff-0f1be8eec416e71c28b41efb0544df26 but in current implementation transactions never returned
And current rpc expecting `/node` to be added: soruce https://mynano.ninja/api and new rpc https://nanoverse.io/api/#account_history
```curl
  https://nanoverse.io/api/node/
-d '{
  "action": "account_history",
  "account": "nano_3jwrszth46rk1mu7rmb4rhm54us8yg1gw3ipodftqtikf5yqdyr7471nsg1k",
  "count": "5"
}'
```
But should be 
```curl
  https://nanoverse.io/api/node
-d '{
  "action": "account_history",
  "account": "nano_3jwrszth46rk1mu7rmb4rhm54us8yg1gw3ipodftqtikf5yqdyr7471nsg1k",
  "count": "5"
}'
```

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -214,7 +214,7 @@ algorand:
 #  key: YOUR_API_KEY
 
 nano:
-  # api: https://mynano.ninja/api
+  api: https://nanoverse.io/api/node
 
 digibyte:
   api: https://dgb1.trezor.io/api
--- a/platform/nano/api.go
+++ b/platform/nano/api.go
@@ -1,7 +1,6 @@
 package nano
 
 import (
-	"encoding/json"
 	"strconv"
 
 	"github.com/trustwallet/blockatlas/coin"
@@ -16,7 +15,6 @@ type Platform struct {
 
 func (p *Platform) Init() error {
 	p.client = Client{blockatlas.InitClient(viper.GetString("nano.api"))}
-	p.client.Headers["Content-Type"] = "application/json"
 	return nil
 }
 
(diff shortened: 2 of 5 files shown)


Pantani:
I did this logic to work with rpc's

https://github.com/trustwallet/blockatlas/blob/master/pkg/blockatlas/jsonrpc.go

config.yml
```diff
@@ -214,7 +214,7 @@ algorand:
 #  key: YOUR_API_KEY
 
 nano:
-  # api: https://mynano.ninja/api
```
kolya182: ninja rpc i paid rpc, replacing with public one that is  free

hewigovens:
Have you tested new account? Should return empty array

platform/nano/api.go
```diff
@@ -30,17 +29,7 @@ func (p *Platform) GetTxsByAddress(address string) (blockatlas.TxPage, error) {
 		return nil, err
 	}
 
-	b, err := json.Marshal(history.History)
```
hewigovens: why remove this? `history` field might be empty string (not array)

kolya182: Ahh i see, looks like NANO RPC bug or stupid decision return string when array normally returned of transactions. Logic was returned, i left a comment to avoid such as misunderstanding.

PR #615, opened 2019-12-06
Refactor pricing api

dpereskokov:


marketdata/market/compound/compound.go
```diff
@@ -16,12 +15,12 @@ type Market struct {
...
 
-func InitMarket() market.Provider {
+func InitMarket(api string, updateTime string) market.Provider {
 	m := &Market{
 		Market: market.Market{
 			Id:         compound,
-			Request:    blockatlas.InitClient(viper.GetString("market.compound.api")),
-			UpdateTime: viper.GetString("market.quote_update_time"),
+			Request:    blockatlas.InitClient(api),
+			UpdateTime: updateTime,
```
vikmeup: `UpdateTime` - does it needs to be int or string?

dpereskokov: It needs to be string since this value is consumed by Cron which is also string:
https://github.com/trustwallet/blockatlas/blob/4ff3ffc9b8e3fac4ceddb33b48be17fb8d833953/marketdata/worker.go#L51

Issue #614, opened 2019-12-05
Add Harmony platform


vikmeup:

vikmeup:
@lorenzozanotto, this will help https://developer.trustwallet.com/blockatlas/newblockchain

lorenzozanotto:
@vikmeup code and tests written, I'm waiting for the Harmony team to unlock the task by providing support on why transactions don't show up using their RPC methods. In the meanwhile I'm going to create a pull request so the code that's been written can be visible. This will allow us to push one button to send the code through the pipeline when Harmony team will solve the matter

PR #624, opened 2019-12-09
Add Harmony support

lorenzozanotto:
Implements #614 

**Don' merge yet**, I'm waiting for the Harmony team to tell me when the transactions could be retrieved successfully via their RPC interface. I would like to have the code reviewed though, so once they're ready, we can push a button to have the feature in production

**What has been done?**

- Add transactions retrievement by address
- Add current block number retrievement
- Add block retrievement by number
- The method `RPCCall` accepts the `params` parameter as `interface{}` since Harmony requires to pass an object to the classic array


pkg/blockatlas/jsonrpc.go
```diff
@@ -41,7 +41,9 @@ func (r *RpcResponse) GetObject(toType interface{}) error {
 	return nil
 }
 
-func (r *Request) RpcCall(result interface{}, method string, params []string) error {
+func (r *Request) RpcCall(result interface{}, method string, params interface{}) error {
+	r.Headers["Content-Type"] = "application/json"
```
Pantani: You can break other implementations. This method is used by all json-rpc calls inside Blockatlas. You can set a custom header inside the platform implementation

platform/harmony/api.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/trustwallet/blockatlas/pkg/blockatlas"
+	"strconv"
+)
+
+type Platform struct {
+	client Client
+}
+
+func (p *Platform) Init() error {
+	p.client = Client{blockatlas.InitClient(viper.GetString("harmony.rpc"))}
```
Pantani:
```suggestion
	p.client = Client{blockatlas.InitClient(viper.GetString("harmony.rpc"))}
	c.Headers["Content-Type"] = "application/json"
```

platform/harmony/api.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+func (p *Platform) Coin() coin.Coin {
+	return coin.Coins[coin.ONE]
+}
+
+func (p *Platform) GetTxsByAddress(address string) (blockatlas.TxPage, error) {
+	result, err := p.client.GetTxsOfAddress(address)
+	if err != nil {
+		return []blockatlas.Tx{}, err
```
Pantani:
```suggestion
		return blockatlas.TxPage{}, err
```

platform/harmony/api.go
```diff
@@ -0,0 +1,99 @@
...
+	return p.client.CurrentBlockNumber()
+}
+
+func (p *Platform) GetBlockByNumber(num int64) (*blockatlas.Block, error) {
+	if srcBlock, err := p.client.GetBlockByNumber(num); err == nil {
+		block := p.NormalizeBlock(&srcBlock)
+		return &block, nil
+	} else {
+		return nil, err
+	}
```
Pantani: You don't need this else

```
	if srcBlock, err := p.client.GetBlockByNumber(num); err == nil {
		block := p.NormalizeBlock(&srcBlock)
		return &block, nil
	}
	return nil, err
```

platform/harmony/api.go
```diff
@@ -0,0 +1,99 @@
...
+		return blockatlas.Block{}
+	}
+	return blockatlas.Block{
+		ID:     block.Hash,
+		Number: int64(blockNumber),
+		Txs:    NormalizeTxs(block.Transactions),
+	}
+}
+
+func NormalizeTxs(txs []Transaction) []blockatlas.Tx {
```
Pantani:
```suggestion
func NormalizeTxs(txs []Transaction) blockatlas.TxPage {
```

Pantani: only a suggestion to be more clean

platform/harmony/api.go
```diff
@@ -0,0 +1,99 @@
...
+	normalizeTxs := make([]blockatlas.Tx, 0)
```
Pantani:
```suggestion
	normalizeTxs := make(blockatlas.TxPage, 0)
```

platform/harmony/api.go
```diff
@@ -0,0 +1,99 @@
...
+	for _, srcTx := range txs {
+		normalized, isCorrect := NormalizeTx(&srcTx)
+		if !isCorrect {
+			return []blockatlas.Tx{}
```
Pantani:
```suggestion
			return blockatlas.TxPage{}
```

platform/harmony/api.go
```diff
@@ -0,0 +1,99 @@
...
+		}
+		normalizeTxs = append(normalizeTxs, normalized)
+	}
+	return normalizeTxs
+}
+
+func NormalizeTx(trx *Transaction) (tx blockatlas.Tx, b bool) {
+	gasPrice, err := hexToInt(trx.GasPrice)
```
Pantani: You need to handle this error
```
if err != nil {
	return blockatlas.Tx{}, false
}
```

platform/harmony/api.go
```diff
@@ -0,0 +1,99 @@
...
+	gas, err := hexToInt(trx.Gas)
```
Pantani: You need to handle this error
```
if err != nil {
	return blockatlas.Tx{}, false
}
```

platform/harmony/api.go
```diff
@@ -0,0 +1,99 @@
...
+	fee := gas * gasPrice
+	literalFee := strconv.Itoa(int(fee))
+
+	value, err := hexToInt(trx.Value)
```
Pantani: You need to handle this error
```
if err != nil {
	return blockatlas.Tx{}, false
}
```

platform/harmony/api_test.go
```diff
@@ -0,0 +1,77 @@
...
+	Block:    18,
+	Status:   blockatlas.StatusCompleted,
+	Meta: blockatlas.Transfer{
+		Value:    "100000000000000000",
+		Symbol:   "ONE",
+		Decimals: 18,
+	},
+}
+
+func TestNormalize(t *testing.T) {
```
Pantani: use this package to make the unit tests
"github.com/stretchr/testify/assert"

You can see a example here
https://github.com/trustwallet/blockatlas/blob/master/platform/binance/api_test.go#L383

platform/harmony/client.go
```diff
@@ -0,0 +1,44 @@
...
+	return
+}
+
+func (c *Client) CurrentBlockNumber() (int64, error) {
+	var nodeInfo string
+	err := c.RpcCall(&nodeInfo, "hmy_blockNumber", nil)
+	if err != nil {
+		return 0, err
+	}
+	decimalBlock, _ := hexToInt(nodeInfo)
```
Pantani:
```suggestion
	decimalBlock, err := hexToInt(nodeInfo)
	if err != nil {
		return 0, err
	}
```

You need to handle this error

platform/harmony/client.go
```diff
@@ -0,0 +1,44 @@
...
+
+type Client struct {
+	blockatlas.Request
+}
+
+func (c *Client) GetTxsOfAddress(address string) (txPage *TxResult, err error) {
+	params := []interface{}{
+		map[string] interface{}{"address": address, "fullTx": true},
+	}
+	err = c.RpcCall(&txPage, "hmy_getTransactionsHistory", params)
```
Pantani: Use struct
```
	type params struct {
		address string `json:"address"`
		fullTx  bool       `json:"fullTx"`
	}
	err = c.RpcCall(&txPage, "hmy_getTransactionsHistory",  params{address: address, fullTx: true})
```

config.yml
```diff
@@ -220,3 +220,7 @@ nano:
 
 digibyte:
   api: https://dgb1.trezor.io/api
+
+harmony:
+  rpc: https://harmony-rpc.trustwalletapp.com
```
vikmeup: keep one, use public one

platform/harmony/api.go
lorenzozanotto: I tried to do so, but it seems like that an assignment inside a conditional will make the assigned values live only in the conditional scope. If I write `err` outside the conditional I will get it as `Unresolved reference`, that's why I found the else way to be helpful in that case.

Another way could be declaring a `var err error` before the conditional like so

```go
var err error
if srcBlock, err := p.client.GetBlockByNumber(num); err == nil {
    block := p.NormalizeBlock(&srcBlock)
    return &block, nil
}
return nil, err
```

platform/harmony/api.go
lorenzozanotto: Gold

platform/harmony/api.go
lorenzozanotto: In the `NormalizeTx` function there are a lot of `err` assignments, there's a way to avoid 

```go
if err != nil {
    return blockatlas.Tx{}, false
}
```

after each line? Otherwise I'll go with duplication

platform/harmony/api_test.go
lorenzozanotto: Nice! Took as a reference other coins that had a different testing approach. I'll refactor my tests to use the example you provided

platform/harmony/api.go
Pantani: try that

```
srcBlock, err := p.client.GetBlockByNumber(num)
if err != nil {
    return nil, err
}
block := p.NormalizeBlock(&srcBlock)
return &block, nil
```

platform/harmony/api.go
Pantani: Yes, you need to handle all errors, it's annoying, I know, but is the Go pattern, you always need to handle an error, and try to put all possible information inside the error, we have our own error package to do that `github.com/trustwallet/blockatlas/pkg/errors`

Example:
https://github.com/trustwallet/blockatlas/blob/ee5e423df6adf6b431b2aa0460d8becb95d6f53b/pkg/blockatlas/client.go#L63

You can push this error to sentry if you add the method in the end:`.PushToSentry()`

lorenzozanotto: Great! Thank you for the explanation, if that's the correct way I'll do this without further ado

platform/harmony/model.go
```diff
@@ -0,0 +1,50 @@
...
+	GasLimit         string        `json:"gasLimit"`
+	GasUsed          string        `json:"gasUsed"`
+	Hash             string        `json:"hash"`
+	LogsBloom        string        `json:"logsBloom"`
+	Miner            string        `json:"miner"`
+	MixHash          string        `json:"mixHash"`
+	Nonce            int           `json:"nonce"`
+	Number           string        `json:"number"`
+	ParentHash       string        `json:"parentHash"`
+	ReceiptsRoot     string        `json:"receiptsRoot"`
```
dpereskokov: lets get rid of unused props

platform/harmony/client.go
lorenzozanotto: @Pantani tried to used structs and for some reason the `fullTx` boolean gets translated as `"true"` and not `true`. Working with a map inside an interface makes it work properly

platform/harmony/model.go
lorenzozanotto: Done!

platform/harmony/model.go
```diff
@@ -15,36 +15,13 @@ type Transaction struct {
 	Gas              string `json:"gas"`
 	GasPrice         string `json:"gasPrice"`
 	Hash             string `json:"hash"`
-	Input            string `json:"input"`
-	Nonce            string `json:"nonce"`
```
vikmeup: you need `nonce` value

lorenzozanotto: Added it to the `Sequence` field

lorenzozanotto:
@Pantani I discovered that `hexToInt(hex string)`, taken from **venchain/api.go** is not working properly. 

The hex string `0x1158e460913d00000` gets converted as int `18446744073709551615` instead of `20000000000000000000`. I'm currently using `util.HexToDecimal(Value)` since I need a string output and that works well. The venchain method, after the HexToDecimal does a `strconv.ParseUint(nonceStr, 10, 64)` which generates an incorrect value in that specific case

PR #625, opened 2019-12-09
SQL Package

Pantani:
Add wrapper package to handle database connections with GORM


pkg/storage/sql/sql.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/jinzhu/gorm"
+	"github.com/trustwallet/blockatlas/pkg/errors"
+	"github.com/trustwallet/blockatlas/pkg/storage/util"
+)
+
+type sql struct {
+	Client *gorm.DB
+}
+
+type Handler func(db *gorm.DB, value interface{}) error
```
vikmeup: can you also add methods to add indexes?

Pantani: whats do you mean?

we can add an indexer in gorm using `gorm:"index:addr"` inside the struct

vikmeup: 👍

PR #635, opened 2019-12-13
remove cosmos limit for validators

Pantani:


platform/cosmos/client.go
```diff
@@ -33,7 +33,6 @@ func (c *Client) GetValidators() (validators Validators, err error) {
 	query := url.Values{
 		"status": {"bonded"},
 		"page":   {strconv.FormatInt(1, 10)},
-		"limit":  {strconv.FormatInt(blockatlas.ValidatorsPerPage, 10)},
```
vikmeup: does this always return all validators?

Pantani: yes, 125 validators

PR #638, opened 2019-12-13
fix failed transactions parse

Pantani:


platform/cosmos/api_test.go
```diff
@@ -232,41 +282,87 @@ var transferDst = blockatlas.Tx{
...
+		Decimals: 6,
+	},
+}
 
-	dstJSON, err := json.Marshal(&dst)
-	if err != nil {
-		t.Fatal(err)
-	}
+func TestNormalize(t *testing.T) {
+	testNormalize(t, "test transfer tx", transferSrc, transferDst)
```
dpereskokov: lets use table tests

Pantani: In this case don't make a lot of sense, because the test is already dynamic. But I did the changes.

Pantani: one good example of table tests changes

https://github.com/trustwallet/backend/pull/271/files#diff-aecf3ca183be94d51ad0c89054705d7d

platform/cosmos/api.go
```diff
@@ -94,11 +94,17 @@ func Normalize(srcTx *Tx) (tx blockatlas.Tx, ok bool) {
 		}
 	}
 
+	status := blockatlas.StatusCompleted
+	// https://github.com/cosmos/cosmos-sdk/blob/95ddc242ad024ca78a359a13122dade6f14fd676/types/errors/errors.go#L19
+	if srcTx.Code > 0 {
```
vikmeup: 👍

Issue #612, opened 2019-12-05
Add pricing API support for coingecko 


dpereskokov:

vikmeup:
platforms:

```
["vechain", "neo", "tron", "lisk", "tezos", "stratis", "stellar", "polkadot", "qtum", "waves", "ethereum", "yocoin", "cosmos", "ontology", "binancecoin", "bitshares", "factom", "komodo", "nxt", "counterparty", "eos", "omni", "dash", "bitcoin-cash", "openledger", "rootstock", "gochain", "aion", "wanchain", "ardor", "nem", "icon", "metaverse-etp", "ethereum-classic"]
```

https://api.coingecko.com/api/v3/coins/list?include_platform=true

PR #648, opened 2019-12-16
Add pricing API support for coingecko

dpereskokov:
Fixes: #612


marketdata/market/coingecko/coingecko_test.go
```diff
@@ -0,0 +1,104 @@
...
+				},
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			gotTickers := m.normalizeTickers(tt.args.prices, tt.args.provider)
+			now := time.Now()
+			sort.Slice(gotTickers, func(i, j int) bool {
+				gotTickers[i].LastUpdate = now
```
vikmeup: why do we do sorting in every platform for pricing? how could this be improved?

dpereskokov: We generate platforms mapping dynamically based on this endpoint: https://api.coingecko.com/api/v3/coins/list?include_platform=true
So I see no need to statically map it as it's been done in cmc. We map prices and coins by CG id.

dpereskokov: @vikmeup is's done only for tests to make sure assert.Equal will compare arrays properly.

marketdata/coingecko/cache.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/trustwallet/blockatlas/pkg/blockatlas"
+	"github.com/trustwallet/blockatlas/pkg/errors"
+)
+
+type Cache map[string][]CoinResult
+
+func (c Cache) GetCoinsBySymbol(id string) (coins []CoinResult, err error) {
+	coins, ok := c[id]
+	if !ok {
+		err = errors.E("No coin found by id", errors.Params{"id": id}).Err
```
Pantani:
```suggestion
		err = errors.E("No coin found by id", errors.Params{"id": id})
```

marketdata/coingecko/client.go
```diff
@@ -0,0 +1,82 @@
...
+				logger.Error(err)
+				return
+			}
+			prChan <- cp
+		}(i)
+
+		i += bucketSize
+	}
+
+	go func() {
```
Pantani: see Fan-out, fan-in in the official Go docs

https://blog.golang.org/pipelines

marketdata/market/coingecko/coingecko.go
```diff
@@ -0,0 +1,94 @@
...
+		client: coingecko.NewClient(api),
+		Market: market.Market{
+			Id:         id,
+			UpdateTime: updateTime,
+		},
+	}
+	return m
+}
+
+func (m *Market) GetData() (result blockatlas.Tickers, err error) {
```
Pantani:
```suggestion
func (m *Market) GetData() (blockatlas.Tickers, error) {
```

marketdata/coingecko/client.go
```diff
@@ -0,0 +1,82 @@
...
+
+func (c *Client) FetchCoinsList() (coins GeckoCoins, err error) {
+	values := url.Values{
+		"include_platform": {"true"},
+	}
+	err = c.GetWithCache(&coins, "v3/coins/list", values, time.Hour)
+	return
+}
+
+func coinIds(coins GeckoCoins) []string {
```
Pantani:
```suggestion
func (coins GeckoCoins)coinIds() []string {
```

marketdata/coingecko/client.go
```diff
@@ -0,0 +1,82 @@
...
+	c := Client{
+		Request: blockatlas.InitClient(api),
+	}
+	return &c
+}
+
+func (c *Client) FetchLatestRates(coins GeckoCoins) (prices CoinPrices) {
+	ci := coinIds(coins)
+
+	bucketSize := 500
```
Pantani: can you turn it in a constant?

marketdata/market/coingecko/coingecko.go
dpereskokov: This doesn't work here as I use named return statement

Pantani: Don't mix the implicit return variables and the standard return is confusing

You are using only to handle the error, don't make sense create a variable global for all method to use only if have an error. Because you are not using that in the [line 40](https://github.com/trustwallet/blockatlas/pull/648/files#diff-66d5ab40d52392088915d9460e055f2cR40)

this implicit return variable works well if you need to unmarshal the value inside the variables or for append slice or map

vikmeup:
What you provide list of assets for testing? That only pulled from CG

marketdata/market/coingecko/coingecko.go
dpereskokov: fixed [line 40](https://github.com/trustwallet/blockatlas/pull/648/files#diff-66d5ab40d52392088915d9460e055f2cR40)

Issue #617, opened 2019-12-07
[Cosmos] Add GetReward transaction type


vikmeup:
Ref: https://github.com/trustwallet/blockatlas/blob/master/platform/cosmos/model.go#L151, https://github.com/trustwallet/blockatlas/blob/master/platform/cosmos/model.go#L16

Transaction example: https://www.mintscan.io/txs/789b9a80f9506f3490f4d51559448692084ba56d88829046c83311b239b70ffd

PR #650, opened 2019-12-16
fix cosmos stake direction

Pantani:
`/v1/cosmos/cosmos1dx27g0kzhwej0ekcf2k9hsktcxnmpl7fcehcvq`

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -91,6 +91,9 @@ func makeTxRoute(router gin.IRouter, api blockatlas.Platform, path string) {
 
 		page := make(blockatlas.TxPage, 0)
 		for _, tx := range txs {
+			if tx.Direction != "" {
+				goto AddTx
+			}
 			tx.Direction = blockatlas.DirectionOutgoing
 			if tx.To == address {
 				tx.Direction = blockatlas.DirectionIncoming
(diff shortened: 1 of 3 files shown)


api/handlers.go
```diff
@@ -91,13 +91,17 @@ func makeTxRoute(router gin.IRouter, api blockatlas.Platform, path string) {
 
 		page := make(blockatlas.TxPage, 0)
 		for _, tx := range txs {
+			if tx.Direction != "" {
+				goto AddTx
```
vikmeup: is it safe to use this? maybe consider if else instead?

platform/cosmos/api.go
```diff
@@ -158,8 +158,10 @@ func fillDelegate(tx *blockatlas.Tx, delegate MessageValueDelegate, msgType stri
 	title := blockatlas.KeyTitle("")
 	switch msgType {
 	case MsgDelegate:
+		tx.Direction = blockatlas.DirectionOutgoing
 		title = blockatlas.AnyActionDelegation
```
vikmeup: can you also add this: https://github.com/trustwallet/blockatlas/issues/617, should be a quick one

api/handlers.go
Pantani: yess!!! `goto` method is normal and safe in Go, the golang compiler cannot compile the app if the `goto` skip handling some error or instantiate a variable. you can only use `goto` in safe places

platform/cosmos/api.go
Pantani: this is not so easy because we need to filter 2 log messages, I going to do that in another PR. And keep this PR simple

PR #652, opened 2019-12-16
run go fmt for harmony platform

Pantani:


Pantani:
@vikmeup  the Goland can automatically do that. but we need to configure the IDE. Or we can run `make fmt` from CI and commit the changes

Issue #541, opened 2019-11-17
Integration tests for naming service


hewigovens:
test .eth/.zil, and .crypto in the future

hewigovens:
what's the status of this issue?

EnoRage:
@vikmeup please, provide information about it. Maybe open another issue for it and close that?

vikmeup:
We need to test endpoint and it's result (JSON response):

Example:
Request: https://blockatlas.trustwalletapp.com/ns/lookup?coin=60&name=vikmeup.eth
Response: 
```
{
   "result":"0x84E79D544B4b13bC3560069cfD56A9D5bbE7521d",
   "coin":60
}
```

PR #656, opened 2019-12-18
add integration tests fixtures for post and query string

Pantani:
closes https://github.com/trustwallet/blockatlas/issues/541

Now we can add a body or a query string to do request in the integration tests

- `testdata` - Folder contains the JSON fixtures
- `body_fixtures.json` - Body fixtures for POST requests
- `coin_fixtures.json` - Coin fixtures for path parameters
- `query_fixtures.json` - Query string for GET requests
- `exclude.json` - API's need to be excluded from integration tests


pkg/integration/testdata/query_fixtures.json
```diff
@@ -0,0 +1,10 @@
+{
+  "/ns/lookup": {
```
hewigovens: we only test http 200 code?

vikmeup: would be test to test response result

vikmeup:
In order to complete https://github.com/trustwallet/blockatlas/issues/541, we need to test: .ens, .crypto, .zil domain names, for ethereum and other coins, for both existing and non existing names and make results are correct

Pantani:
> In order to complete #541, we need to test: `.ens`, `.crypto`, `.zil` domain names, for ethereum and other coins, for both existing and non existing names and make results are correct

@vikmeup Done!!!

vikmeup:
@Pantani as I understand those tests only check for 200, it does not tests what it needs to. 

https://github.com/trustwallet/blockatlas/issues/541 need to test different services for their responses data, not just the status code.

PR #659, opened 2019-12-19
Add Kava coin support

vikmeup:
Work in progress

- [x] Change to use `p.CoinIndex`
- [x] Update tests


platform/cosmos/model.go
```diff
@@ -78,7 +80,7 @@ func (a Attributes) GetWithdrawRewardValue() int64 {
 	result := int64(0)
 	for _, att := range a {
 		if att.Key == AttributeAmount {
-			value := strings.Replace(att.Value, UndelegateDenom, "", -1)
+			value := att.Value[:strings.IndexByte(att.Value, 'u')]
```
vikmeup: all values on cosmos seems to have `u` identified in the beginning, since I can't easily inject `platform.Denom`, it might be easier to split it this way

platform/cosmos/api.go
```diff
@@ -11,16 +12,21 @@ import (
...
 
 type Platform struct {
-	client Client
+	client    Client
+	CoinIndex uint
 }
 
 func (p *Platform) Init() error {
-	p.client = Client{blockatlas.InitClient(viper.GetString("cosmos.api"))}
+	p.client = Client{blockatlas.InitClient(viper.GetString(p.ConfigKey()))}
```
vikmeup: Same pattern as btc, created an issue to improve this: https://github.com/trustwallet/blockatlas/issues/668

PR #663, opened 2019-12-20
Charts API

dpereskokov:
**Sumary**
Implemented multi provider charts api
Clients are separated from providers structs to reuse same rpc calls

**Usage example**
`/v1/market/charts?currency=USD&timeStart=1574483028&coin=60&token=0xdAC17F958D2ee523a2206206994597C13D831ec7`
token is optional

Fixes: trustwallet/backend#52
Sorry for the big PR.


api/maketdata.go
```diff
@@ -114,3 +115,42 @@ func getTickersHandler(storage storage.Market) func(c *gin.Context) {
...
+// @Accept json
+// @Produce json
+// @Tags charts
+// @Param coin query int true "coin id"
+// @Param token query string false "token id"
+// @Param currency query string false "the currency to show charts" default(USD)
+// @Success 200 {object} blockatlas.ChartData
+// @Router /v1/market/charts [get]
+func getChartsHandler(router gin.IRouter) {
+	var charts = marketdata.InitCharts()
```
vikmeup: is there a way to init once and not every time you call it?

api/maketdata.go
```diff
@@ -114,3 +115,42 @@ func getTickersHandler(storage storage.Market) func(c *gin.Context) {
...
+		}
+		token := c.Query("token")
+
+		days, err := strconv.Atoi(c.Query("days"))
+		if err != nil {
+			ginutils.RenderError(c, http.StatusInternalServerError, "Invalid days count")
+			return
+		}
+
+		currency := c.DefaultQuery("currency", blockatlas.DefaultCurrency)
```
vikmeup: I think we should reject not supported currencies

api/maketdata.go
dpereskokov: It is closure. It is only called once when getChartsHandler() is called

api/maketdata.go
dpereskokov: It is used for the case when you haven't passed `currency=USD` as parameter. Just for request simplification sake.

api/routes.go
```diff
@@ -53,6 +53,8 @@ func LoadPlatforms(root gin.IRouter) {
 	makeCategoriesBatchRoute(v2)
 	makeStakingDelegationsBatchRoute(v2)
 	makeStakingDelegationsSimpleBatchRoute(v2)
+
+	getChartsHandler(v1)
```
vikmeup: should we be consistent and call it `makeChartsRoute` ?

kolya182:
Can you add description how much paid request implementation consumes, does it use free endpoints at all, what are the constraint here (like using public api from cmc) etc...

Pantani:
try to do small PR next time, you can split the clients first and implement the charts after

Pantani:
nice code. Good work @dpereskokov !!! 👏👏

PR #669, opened 2019-12-23
Add GetMaxAPR for Cosmos

vikmeup:
@Pantani would you be able to add caching for this? we should just call this once a week - maybe store in redis? 

this is an example how it should work, please take this from here.

### DIFF
--- a/platform/cosmos/stake.go
+++ b/platform/cosmos/stake.go
@@ -10,6 +10,10 @@ import (
 	"time"
 )
 
+const (
+	defaultAnnualReward = 8.1
+)
+
 func (p *Platform) GetValidators() (blockatlas.ValidatorPage, error) {
 	results := make(blockatlas.ValidatorPage, 0)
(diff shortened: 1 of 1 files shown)


platform/cosmos/stake.go
```diff
@@ -38,15 +42,33 @@ func (p *Platform) GetValidators() (blockatlas.ValidatorPage, error) {
...
+			Annual: p.GetMaxAPR(),
+		},
 		MinimumAmount: blockatlas.Amount("0"),
 		LockTime:      1814400,
 		Type:          blockatlas.DelegationTypeDelegate,
 	}
 }
 
+func (p *Platform) GetMaxAPR() float64 {
+	validators, err := p.GetValidators()
```
Pantani: I think this API is going to be slow. Can we try to cache something? like the `GePool` or `GetInflation`request

vikmeup: Take a look at my comment in the PR

Issue #683, opened 2020-01-03
Bitcoincash: cannot unmarshal number -1 into Go struct field Transaction.transactions.blockHeight of type uint64


kolya182:
## Summary of Bug
Expected zero transactions returned for address with only one transaction is still in pending

```js
{
error: "json: cannot unmarshal number -1 into Go struct field Transaction.transactions.blockHeight of type uint64",
meta: {
method: "GET",
url:"http://localhost:8420/v2/xpub/xpub6Bm1ZCno7WctsAuVr1ADB8EX23cocDzXNYsi9FkaQJtr74CZaYhT5AnAedj78BW9La6JNVkwFPY7VYpsJAL8y1DEq7mB2w6sEjoBsZdZRrP?details=txs&pageSize=100&tokens=derived"
},
type: "Platform Unmarshal Error"
}
```
## Steps to Reproduce

http://localhost:8420/v1/bitcoincash/xpub/xpub6Bm1ZCno7WctsAuVr1ADB8EX23cocDzXNYsi9FkaQJtr74CZaYhT5AnAedj78BW9La6JNVkwFPY7VYpsJAL8y1DEq7mB2w6sEjoBsZdZRrP

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #684, opened 2020-01-03
Handle pending transaction

kolya182:
Fixes https://github.com/trustwallet/blockatlas/issues/683

Now pending transactions returned

### DIFF
--- a/platform/bitcoin/api.go
+++ b/platform/bitcoin/api.go
@@ -197,7 +197,7 @@ func NormalizeTransaction(transaction *Transaction, coinIndex uint) blockatlas.T
 	}
 
 	status := blockatlas.StatusCompleted
-	if transaction.Confirmations == 0 || transaction.BlockHeight == -1 {
+	if transaction.Confirmations == 0 {
 		status = blockatlas.StatusPending
 	}
 
--- a/platform/bitcoin/api.go
+++ b/platform/bitcoin/api.go
@@ -197,7 +197,7 @@ func NormalizeTransaction(transaction *Transaction, coinIndex uint) blockatlas.T
 	}
 
 	status := blockatlas.StatusCompleted
-	if transaction.Confirmations == 0 {
+	if transaction.Confirmations == 0 || transaction.BlockHeight == -1 {
 		status = blockatlas.StatusPending
 	}
 
@@ -211,7 +211,7 @@ func NormalizeTransaction(transaction *Transaction, coinIndex uint) blockatlas.T
 		Fee:      blockatlas.Amount(transaction.Fees),
 		Date:     int64(transaction.BlockTime),
 		Type:     blockatlas.TxTransfer,
-		Block:    transaction.BlockHeight,
+		Block:    uint64(transaction.BlockHeight),
 		Status:   status,
 		Sequence: 0,
 		Meta: blockatlas.Transfer{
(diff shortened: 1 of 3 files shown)


platform/bitcoin/model.go
```diff
@@ -27,7 +27,7 @@ type Transaction struct {
 	Vin           []Output `json:"vin"`
 	Vout          []Output `json:"vout"`
 	BlockHash     string   `json:"blockHash"`
-	BlockHeight   uint64   `json:"blockHeight"`
+	BlockHeight   int64    `json:"blockHeight"`
```
hewigovens: why int64? negative block number?

platform/bitcoin/api.go
```diff
@@ -305,3 +300,11 @@ func (p *Platform) InferValue(tx *blockatlas.Tx, direction blockatlas.Direction,
 	}
 	return value
 }
+
+func (trx *Transaction) getTransactionStatus() blockatlas.Status {
+	if trx.Confirmations == 0 {
+		return blockatlas.StatusPending
+	} else {
+		return blockatlas.StatusCompleted
```
vikmeup: why `else`?

just `return blockatlas.StatusCompleted`

platform/bitcoin/api_test.go
```diff
@@ -262,3 +263,27 @@ func TestInferDirection(t *testing.T) {
 		}
 	}
 }
+
+func TestTransactionStatus(t *testing.T)  {
+	var trxPending = `{"confirmations": 0}`
+	var trxCompleted = `{"confirmations":1}`
+
```
vikmeup: why? 

just use `Transaction{Confirmations: 1}`

platform/bitcoin/api.go
```diff
@@ -305,3 +300,11 @@ func (p *Platform) InferValue(tx *blockatlas.Tx, direction blockatlas.Direction,
 	}
 	return value
 }
+
+func (trx *Transaction) getTransactionStatus() blockatlas.Status {
```
vikmeup: `getTransactionStatus ` => `Status` or `GetStatus`

platform/bitcoin/model.go
kolya182: Yes, when transaction is pending block hight = `-1`

platform/bitcoin/api.go
hewigovens: in this case, we should check block numbers negative here?

kolya182: I has this check https://github.com/trustwallet/blockatlas/pull/684/commits/3b69f7f9d44c24a25b743c544d0681a751aa6054#diff-7bc94df832258a11b2a6d0a76bf76c87R200 but removed, one condition should be sufficient

platform/bitcoin/api.go
```diff
@@ -305,3 +300,10 @@ func (p *Platform) InferValue(tx *blockatlas.Tx, direction blockatlas.Direction,
 	}
 	return value
 }
+
+func (trx *Transaction) getStatus() blockatlas.Status {
```
vikmeup: `trx` = `tx`

platform/bitcoin/api_test.go
```diff
@@ -262,3 +263,17 @@ func TestInferDirection(t *testing.T) {
 		}
 	}
 }
+
+func TestTransactionStatus(t *testing.T) {
+	var tests = []struct {
+		Trx      Transaction
```
vikmeup: `Trx` = `Tx`

kolya182:
Test fails due to slow response from aion explorer, reported to the team to fix it, merging for now

PR #688, opened 2020-01-05
Remove the need of having redis for developing

### LINKED ISSUE #667: Allow to run blockatlas without reddis
## Summary of Bug

Currently if I try to run blockatlas without Redis, it won't lunch, it would be great to launch without reddis - but in this case some functionality will be disable - which is ok

```
time="2019-12-22T13:16:15-08:00" level=fatal msg="{\"error\":\"dial tcp [::1]:6379: connect: connection refused: Redis connection test failed\",\"meta\":{}}"
```

## Steps to Reproduce

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

fionera:
This will fix #667 by adding an interface for the cache backend and then implementing a mock structure which doesnt save and never returns anything.


cmd/root.go
```diff
@@ -22,14 +25,20 @@ var (
 			confPath, _ := cmd.Flags().GetString("config")
 			config.LoadConfig(confPath)
 
+			// Are we in dev mode?
+			DevMode, _ = cmd.Flags().GetBool("dev")
```
Pantani: start with lowercase for local variables

storage/storage.go
```diff
@@ -2,16 +2,24 @@ package storage
...
 
-func New() *Storage {
+func New(useMock bool) *Storage {
 	s := new(Storage)
+	if useMock {
+		s.Backend = &mock.Mock{}
+	} else {
+		s.Backend = &redis.Redis{}
+	}
+
```
Pantani:
```
	s.Backend = &mock.Mock{}
	if !useMock {
		s.Backend = &redis.Redis{}
	}
```

cmd/root.go
```diff
@@ -1,19 +1,22 @@
...
 	"github.com/trustwallet/blockatlas/storage"
-	"os"
-	"os/signal"
-	"syscall"
 )
 
 var (
-	Storage = storage.New()
+	Storage *storage.Storage
+	DevMode bool
```
Pantani: why did you use global var?

fionera: for Storage or DevMode? DevMode just because it is in the rootCmd. I can change it if you want ^^

cmd/root.go
fionera: I intentionally made it as global variable since it is in the rootCmd

vikmeup:
We can close this one for now, not needed

PR #689, opened 2020-01-05
k8s: Add example deployment

fionera:


k8s/deployment.yml
```diff
@@ -0,0 +1,30 @@
...
+    spec:
+      containers:
+        - env:
+            - name: ATLAS_MARKET_ENABLED
+              value: "true"
+            - name: ATLAS_OBSERVER_ENABLED
+              value: "true"
+            - name: ATLAS_STORAGE_REDIS
+              value: redis://redis-ha:6379 # I can recommend the Redis Helm Chart https://github.com/helm/charts/tree/master/stable/redis
+          image: fionera/blockatlas
```
vikmeup: can we use `trustwallet/blockatlas` ?

fionera: Oh yeah forgot to change that :)

PR #700, opened 2020-01-08
[NewTxType] for multiple currency chains

YunSuk-Yeo:
Multiple currency chains usually pay fee and send with multiple currency

* Append TxMultiCurrency & TxMultiCurrencyAnyAction type

PR #701, opened 2020-01-08
[Terra] Integration

YunSuk-Yeo:
This PR is following #700 

* Integrating block apis for Terra


vikmeup:
We currently do not have plans to support and terra blockchain. We might change this decision later.

PR #709, opened 2020-01-14
[Market] fix pull charts for all times in markets #708

### LINKED ISSUE #708: [Market] Not able to pull latest charts for all time
## Summary of Bug

Not able to pull charts for all times

## Steps to Reproduce

GET `/v1/market/charts?currency=USD&time_start=1&coin=60&token=`

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

prazd:
The problem was in checking the result of the function - 
if error == nil, but the result is empty, then it returned
I also decided to make the passage through providers concurrently

### DIFF
--- a/marketdata/charts.go
+++ b/marketdata/charts.go
@@ -7,6 +7,7 @@ import (
 	cmc "github.com/trustwallet/blockatlas/marketdata/chart/coinmarketcap"
 	"github.com/trustwallet/blockatlas/pkg/blockatlas"
 	"github.com/trustwallet/blockatlas/pkg/errors"
+	"sync/atomic"
 )
 
 type Charts struct {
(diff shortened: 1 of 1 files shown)


marketdata/charts.go
```diff
@@ -28,14 +29,34 @@ func InitCharts() *Charts {
...
+	errChan := make(chan struct{})
+
+	var errCount int32
+	for _, provider := range c.chartProviders {
+		go func(provider chart.Provider) {
+			charts, err := provider.GetChartData(coin, token, currency, timeStart)
+			if err != nil || len(charts.Prices) == 0 {
+				if int(atomic.LoadInt32(&errCount)) == len(c.chartProviders) - 1 {
+					errChan <- struct{}{}
+					return
```
dpereskokov: I would remove this return and use `>=` instead of `==` mark

marketdata/charts.go
```diff
@@ -28,14 +29,34 @@ func InitCharts() *Charts {
...
-		return charts, nil
+	chartsDataChan := make(chan blockatlas.ChartData, len(c.chartProviders))
+	errChan := make(chan struct{})
+
+	var errCount int32
+	for _, provider := range c.chartProviders {
+		go func(provider chart.Provider) {
+			charts, err := provider.GetChartData(coin, token, currency, timeStart)
+			if err != nil || len(charts.Prices) == 0 {
+				if int(atomic.LoadInt32(&errCount)) == len(c.chartProviders) - 1 {
```
dpereskokov: why do you use `len(providers) - 1`? It means if you have 5 providers and 4 of them fail you will send error to err chan then it will be closed and if 5th provider also fails you will send error into closed errors channel and app will be crashed.

dpereskokov:
I would rather keep it simple. Too complicated logic here.

marketdata/charts.go
prazd: if we have 5 providers, and at this iteration errCount == 4 (len(providers) - 1), it means that this provider is the last (5) and value will send to the errChan

marketdata/charts.go
prazd: I don't think so - we need to send the value to the error channel only when all the providers did not work correctly

PR #714, opened 2020-01-15
set minimum timestamp date for charts

Pantani:
# Headline
set minimum timestamp date for charts 

## Issue
`time_start` must be a valid ISO 8601 timestamp or unix time value

closes https://github.com/trustwallet/blockatlas/issues/708

### DIFF
--- a/marketdata/charts.go
+++ b/marketdata/charts.go
@@ -7,6 +7,7 @@ import (
 	cmc "github.com/trustwallet/blockatlas/marketdata/chart/coinmarketcap"
 	"github.com/trustwallet/blockatlas/pkg/blockatlas"
 	"github.com/trustwallet/blockatlas/pkg/errors"
+	"github.com/trustwallet/blockatlas/pkg/numbers"
 	"sort"
 )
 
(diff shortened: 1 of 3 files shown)


vikmeup:
Where is this magic number come from?

Pantani:
> Where is this magic number come from?

`time_start` must be a valid ISO 8601 timestamp or unix time value. If you try to use the CMC api for charts, this is the minimum value accepted is `1000000000`

Wrong:
`https://web-api.coinmarketcap.com/v1/cryptocurrency/quotes/historical?convert=USD&format=chart_crypto_details&id=1027&interval=1d&time_end=1579117918&time_start=999999999`

Right:
`https://web-api.coinmarketcap.com/v1/cryptocurrency/quotes/historical?convert=USD&format=chart_crypto_details&id=1027&interval=1d&time_end=1579117918&time_start=1000000000`

marketdata/charts.go
```diff
@@ -28,6 +32,9 @@ func InitCharts() *Charts {
 
 func (c *Charts) GetChartData(coin uint, token string, currency string, timeStart int64) (blockatlas.ChartData, error) {
 	chartsData := blockatlas.ChartData{}
+	if timeStart < minUnixData {
+		timeStart = minUnixData
```
vikmeup: `max(timeStart, minUnixData)` like this

marketdata/charts.go
```diff
@@ -9,6 +9,10 @@ import (
 	"github.com/trustwallet/blockatlas/pkg/errors"
 )
 
+const (
+	minUnixData = 1000000000
```
vikmeup: rename to `minUnixTime `

marketdata/charts.go
Pantani: Do you mean to create a method to verify the max value?

```
func max(timeStart, min int64) int64 {
	if timeStart < min {
		return min
	}
	return timeStart
}
```

vikmeup: There is a function for `Max()`: https://github.com/trustwallet/blockatlas/blob/master/pkg/numbers/number.go#L10

Issue #643, opened 2019-12-15
Improve collectibles parsing


vikmeup:
1) https://api.opensea.io/collections/?asset_owner=0x84e79d544b4b13bc3560069cfd56a9d5bbe7521d, use `slug` instead of `primary_asset_contracts`
2) https://api.opensea.io/collections/?asset_owner=0x84e79d544b4b13bc3560069cfd56a9d5bbe7521d, from comment: it looks like you're using the contract's description there instead of the collection's description. we can't change the Enjin contract description to that. but the collection description already says that here in the API data: https://api.opensea.io/collections/?asset_owner=0x84e79d544b4b13bc3560069cfd56a9d5bbe7521d

Pantani:
@dpereskokov we need to update the slug query as well before to close this issue

PR #649, opened 2019-12-16
use slug as id for fetching collectibles

dpereskokov:
Fixes: #643 (1)

PR #717, opened 2020-01-16
use slug as id for fetching collectibles

dpereskokov:
Fixes: #643 (1)
Previous PR: #649 

**Description**
Old collectibles ids differ from the new ones. Hence to avoid duplicating collectibles in the list we should move them to separate api version

New api: 
```
/v3/collectibles/categories
/v3/\(server)/collections/\(address)/collection/\(assetAddress)
```

Make sure to clear collectibles local db or do migration since new ids might result in duplications.


platform/ethereum/api.go
```diff
@@ -227,6 +237,58 @@ func NormalizeCollectionPage(collections []Collection, coinIndex uint, owner str
...
+		Id:              collectionId,
+		CategoryAddress: collectionId,
+		Address:         owner,
+		Version:         version,
+		Coin:            coinIndex,
+		Type:            collectionType,
+	}
+}
+
+//TODO: remove once most of the clients will be updated
```
vikmeup: can you set deadline in 2 months

Issue #710, opened 2020-01-14
[Market] Return pricing elements amount based on the time


vikmeup:
## Summary of Bug

Currently `/v1/market/charts?currency=USD&time_start=1&coin=60&token=time_start=1577556083` returns 1643 price values, which is too much.

Would be a better solution to return amount of prices based on the time:

1 hours: 60 prices (each minute)
24 hours (every 15 or 30 minutes)
7 days (every 4 hours)
28 days (every 12 hours)
1 year (every 1 day)
All times (every 1-7 day, depends on existence of prices) 

This is not defined, we might need to adjust this to be consistent across different values.

## Steps to Reproduce

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #721, opened 2020-01-17
[Market] Return pricing elements amount based on the time

dpereskokov:
Fixes: #710


api/maketdata.go
```diff
@@ -144,10 +148,14 @@ func makeChartsRoute(router gin.IRouter) {
 			ginutils.RenderError(c, http.StatusInternalServerError, "Invalid time_start")
 			return
 		}
+		maxItems, err := strconv.Atoi(c.Query("max_items"))
+		if err != nil || maxItems <= 0 {
+			maxItems = defaultMaxChartItems
```
vikmeup: Use Min()

api/maketdata.go
```diff
@@ -144,10 +148,14 @@ func makeChartsRoute(router gin.IRouter) {
...
+		}
 
 		currency := c.DefaultQuery("currency", blockatlas.DefaultCurrency)
 
-		chart, err := charts.GetChartData(uint(coinId), token, currency, timeStart)
```
vikmeup: Add documentation to this parameter

Pantani: run `swag init` after adding this parameter in the documentation

marketdata/charts.go
```diff
@@ -52,8 +53,23 @@ func (c *Charts) GetCoinInfo(coin uint, token string, currency string) (blockatl
...
 
-func normalizePrices(prices []blockatlas.ChartPrice) {
+func normalizePrices(prices []blockatlas.ChartPrice, maxItems int) (result []blockatlas.ChartPrice) {
 	sort.Slice(prices, func(p, q int) bool {
 		return prices[p].Date < prices[q].Date
 	})
+	if len(prices) > maxItems && maxItems > 0 {
+		skip := int(math.Ceil(float64(len(prices) / maxItems)))
+		i := 0
+		for i < len(prices) {
```
Pantani:
```suggestion
		for i := 0; i < len(prices); i += skip + 1 {
```

marketdata/charts.go
```diff
@@ -52,8 +53,23 @@ func (c *Charts) GetCoinInfo(coin uint, token string, currency string) (blockatl
...
+			result = append(result, prices[i])
+			i += skip + 1
+		}
+		lastPrice := prices[len(prices)-1]
+		if len(result) > 0 && lastPrice.Date != result[len(result)-1].Date {
+			result = append(result, lastPrice)
+		}
```
Pantani:
```suggestion
		for i := 0; i < len(prices) || i == len(result)-1; i += skip + 1 {
			result = append(result, prices[i])
		}
```

dpereskokov:
@Pantani Yes we need to decimate some data since Coingecko for example doesn't have ability to do it on api level. In fact this solution is the same as cmc does on api side. We just approximate data.

https://www.coingecko.com/api/documentations/v3#/coins/get_coins__id__market_chart_range

marketdata/charts.go
dpereskokov: this solution might not work since skip may overcome len(prices) and be bigger. Hence condition `i == len(result)-1` won't take place.

vikmeup:
My main concern would be regarding inconsistent timing for showing charts.

Let's say I request 1 months chart, I want to make sure to show graphs every 4 or 8 hours. How to make sure that results always return:

```
0.00 AM - 123
4.00 AM - 123.12
8.00 AM - 123.45
```

In this case maybe make sense to introduce `1h`, `1d`... params so backend could generate those values properly?

dpereskokov:
This solution can't guarantee it. We only can rely on provider it returns data evenly distributed so the timeframe between dates will be the same with potential little distortion.

dpereskokov:
But problem with parameter like 1h, 4h is that Coingecko doesn't provide us ability to choose timeframes. It is said:
image

marketdata/charts.go
Pantani: But you can remove the lines 68 and 71 in one simple for interaction

```
		for i := 0; i < len(prices); i += skip + 1 {
			result = append(result, prices[i])
		}
```

Pantani:
> But problem with parameter like 1h, 4h is that Coingecko doesn't provide us ability to choose timeframes. It is said:
> image

But you can do this manually, do you need to add the skipped candle data in the next or last candle. You can calculate the timestamp of the current skipped candle using the math below, but you need to know your precision::

`newCandleTime := (currentCandleTime - (currentCandleTime % precision)) + (precision / 2.0)`

Precisions:
```{
  "1m": 60,
  "5m": 300,
  "15m": 900,
  "30m": 1800,
  "1h": 3600,
  "6h": 21600,
  "12h": 43200,
  "1d": 86400,
  "3d": 259200,
  "1w": 604800,
  "2w": 1209600
}```

vikmeup:
I think it’s ok as is. Let’s ship it, we can improve later

dpereskokov:
> > But problem with parameter like 1h, 4h is that Coingecko doesn't provide us ability to choose timeframes. It is said:
> > image
> 
> But you can do this manually, do you need to add the skipped candle data in the next or last candle. You can calculate the timestamp of the current skipped candle using the math below, but you need to know your precision:
> 
> `newCandleTime := (currentCandleTime - (currentCandleTime % precision)) + (precision / 2.0)`
> 
> Precisions:
> 
> ```
>   "1m": 60,
>   "5m": 300,
>   "15m": 900,
>   "30m": 1800,
>   "1h": 3600,
>   "6h": 21600,
>   "12h": 43200,
>   "1d": 86400,
>   "3d": 259200,
>   "1w": 604800,
>   "2w": 1209600
> ```

I'm not sure I got your point. How can it solve the problem with restriction of maximum amount of date points? For example we don't want to send more than 30 points for time range of 1 day in Coingecko? It returns minute date frames. So we will have 60 * 24 points in the returned array. How this math helps us to decimate data to 30 points?

PR #723, opened 2020-01-17
fix market API initialization

Pantani:
# Headline
Fix market API initialization

## Changes
- Use same initialization for all market routes


api/maketdata.go
```diff
@@ -127,10 +131,13 @@ func getTickersHandler(storage storage.Market) func(c *gin.Context) {
 // @Param time_start query int false "Start timestamp" default(1574483028)
 // @Param currency query string false "The currency to show charts" default(USD)
 // @Success 200 {object} blockatlas.ChartData
-// @Router /v1/market/charts [get]
-func makeChartsRoute(router gin.IRouter) {
```
vikmeup: is this intended change?

api/maketdata.go
```diff
@@ -127,10 +131,13 @@ func getTickersHandler(storage storage.Market) func(c *gin.Context) {
...
+// @Router /market/v1/charts [get]
+func getChartsHandler() func(c *gin.Context) {
 	var charts = marketdata.InitCharts()
-	router.GET("/market/charts", func(c *gin.Context) {
+	if charts == nil {
```
dpereskokov: do we need this condition?

api/maketdata.go
```diff
@@ -167,10 +174,13 @@ func makeChartsRoute(router gin.IRouter) {
 // @Param time_start query int false "Start timestamp" default(1574483028)
 // @Param currency query string false "The currency to show coin info in" default(USD)
 // @Success 200 {object} blockatlas.ChartCoinInfo
-// @Router /v1/market/info [get]
-func makeCoinInfoRoute(router gin.IRouter) {
+// @Router /market/v1/info [get]
```
dpereskokov: is this intended change too?

Pantani: no, the old routes were wrong in the swagger file and I made a confusion, fixed now

api/maketdata.go
Pantani: no, the old routes were wrong in the swagger file and I made a confusion, fixed now

api/maketdata.go
```diff
@@ -30,8 +30,12 @@ type Coin struct {
 
 func SetupMarketAPI(router gin.IRouter, db storage.Market) {
 	router.Use(ginutils.TokenAuthMiddleware(viper.GetString("market.auth")))
+	// Ticker
 	router.GET("/ticker", getTickerHandler(db))
 	router.POST("/ticker", getTickersHandler(db))
+	// Charts
+	router.GET("/charts", getChartsHandler())
+	router.GET("/info", getCoinInfoHandler())
```
vikmeup: I would suggest to name as `getCoinInfo`, `getCharts`

PR #726, opened 2020-01-18
Moved common package to blockatlas

dpereskokov:


pkg/numbers/number.go
```diff
@@ -13,3 +22,50 @@ func Max(x, y int64) int64 {
 	}
 	return y
 }
+
+func round(num float64) int {
```
vikmeup: Upper case it to have it exposed

pkg/address/address.go
```diff
@@ -1,10 +1,14 @@
 package address
 
 import (
+	"encoding/hex"
 	"golang.org/x/crypto/sha3"
+	"strconv"
 	"strings"
 )
 
+const addressLength = 40
```
vikmeup: Name it as ethereumAddressLength = 40

PR #731, opened 2020-01-20
make some client methods public

Pantani:
make some client methods public to be reusable

### DIFF
--- a/pkg/blockatlas/client.go
+++ b/pkg/blockatlas/client.go
@@ -89,14 +89,14 @@ func (r *Request) Execute(method string, url string, body io.Reader, result inte
 	return err
 }
 
-func (r *Request) getBase(path string) string {
+func (r *Request) GetBase(path string) string {
 	if path == "" {
 		return fmt.Sprintf("%s", r.BaseUrl)
 	}
 	return fmt.Sprintf("%s/%s", r.BaseUrl, path)
 }
 
-func getBody(body interface{}) (buf io.ReadWriter, err error) {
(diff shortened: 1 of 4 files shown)


pkg/blockatlas/client.go
```diff
@@ -89,14 +89,14 @@ func (r *Request) Execute(method string, url string, body io.Reader, result inte
...
 		return fmt.Sprintf("%s", r.BaseUrl)
 	}
 	return fmt.Sprintf("%s/%s", r.BaseUrl, path)
 }
 
-func getBody(body interface{}) (buf io.ReadWriter, err error) {
+func GetBody(body interface{}) (buf io.ReadWriter, err error) {
 	if body != nil {
 		buf = new(bytes.Buffer)
 		err = json.NewEncoder(buf).Encode(body)
```
dpereskokov: Seems the err check below is redundant

pkg/blockatlas/observer.go
```diff
@@ -0,0 +1,17 @@
+package blockatlas
+
+type Webhook struct {
+	Subscriptions     map[string][]string `json:"subscriptions"`
+	XpubSubscriptions map[string][]string `json:"xpub_subscriptions"`
```
vikmeup: `XpubSubscriptions map[string][]string `json:"xpub_subscriptions"`` let's remove this

Pantani: yes! I going to do that in the next PR, I only need this to finish refactor the notifications inside the backend

Issue #705, opened 2020-01-13
Remove xpub derivation functionality


vikmeup:
## Summary of Bug

Currently we derivate addresses for xpub from the node which is inefficient. We should remove xpub logic and only handle addresses that client provides

## Steps to Reproduce

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

vikmeup:
https://blockchair.com/dogecoin/transaction/133fb7117f9f6ea6fd1cfe411fcb11ad5c99a549aeaf1249b1ae1705d6e8f085


https://blockchair.com/dogecoin/transaction/133fb7117f9f6ea6fd1cfe411fcb11ad5c99a549aeaf1249b1ae1705d6e8f085

Test cases

PR #735, opened 2020-01-21
remove xpub for pushes

Pantani:
remove xpub implementation for push notifications

closes https://github.com/trustwallet/blockatlas/issues/705


Pantani:
> Can you add test cases: [#705 (comment)](https://github.com/trustwallet/blockatlas/issues/705#issuecomment-574445599) in your tests to make sure it send correct push notification webbook

What type of test do you mean? integration tests? because all of the testable methods already have tested, the other methods need database connection and do request to the push services

vikmeup:
unit test, are you sure you have test for change address to yourself? just double cheking

observer/observer.go
```diff
@@ -133,3 +94,15 @@ func getDirection(tx blockatlas.Tx, address string) blockatlas.Direction {
 	}
 	return blockatlas.DirectionOutgoing
 }
+
+func inferUtxoValue(tx *blockatlas.Tx, address string, coinIndex uint) {
+	if len(tx.Inputs) > 0 && len(tx.Outputs) > 0 {
```
vikmeup: maybe return an error here?

Pantani: no, because if is not an utxo tx, the method don't going do to anything

vikmeup: Maybe add property to transaction as type, and only call this function if it’s utxo transaction type?

Issue #718, opened 2020-01-16
Add sorting criteria for validators list


kolya182:
## Summary

Currently list of validators returned in random order

## Problem Definition

Adding sort criteria will help organize validators in the list for better search (Cosmos chain has already 43 validators https://github.com/trustwallet/assets/blob/master/blockchains/cosmos/validators/list.json)  

## Proposal

`Current call`: `/v2/:platform/staking/validators`
`Future call`:     ` /v2/:platform/staking/validators?sort=<sort_criteria>`
`<sort_criteria>` - `"name" | "apr"`

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #736, opened 2020-01-21
Sort validators by APR in desc by default

kolya182:
Closes https://github.com/trustwallet/blockatlas/issues/718
For now just sort in desc by APR as default response, add other sort criteria when needed.
Also do not return validators that disabled on Assets repo, like https://github.com/trustwallet/assets/blob/master/blockchains/tezos/validators/list.json#L82

### DIFF
--- a/services/assets/client.go
+++ b/services/assets/client.go
@@ -4,6 +4,7 @@ import (
 	"github.com/trustwallet/blockatlas/coin"
 	"github.com/trustwallet/blockatlas/pkg/blockatlas"
 	"github.com/trustwallet/blockatlas/pkg/errors"
+	"sort"
 	"time"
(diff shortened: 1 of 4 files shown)


services/assets/client.go
```diff
@@ -55,7 +59,7 @@ func NormalizeValidators(validators []blockatlas.Validator, assets []AssetValida
 	results := make([]blockatlas.StakeValidator, 0)
 	for _, v := range validators {
 		for _, v2 := range assets {
-			if v.ID == v2.ID {
+			if v.ID == v2.ID && v2.Status.Disabled != true {
```
vikmeup: `!v2.Status.Disabled`

PR #743, opened 2020-01-24
[MarketData] Small refactor of SaveRates() method for caching

EnoRage:
change SaveRates(). Now it returns an error if all tokens are not saved in Redis (99% it can be caused only if Redis connection will be lost). It will help to handle the error and change Redis to memory cache (future).

### DIFF
--- a/marketdata/rates.go
+++ b/marketdata/rates.go
@@ -5,7 +5,7 @@ import (
 	"github.com/spf13/viper"
 	"github.com/trustwallet/blockatlas/marketdata/rate"
 	"github.com/trustwallet/blockatlas/marketdata/rate/coingecko"
-	cmc "github.com/trustwallet/blockatlas/marketdata/rate/coinmarketcap"
+	"github.com/trustwallet/blockatlas/marketdata/rate/coinmarketcap"
 	"github.com/trustwallet/blockatlas/marketdata/rate/compound"
 	"github.com/trustwallet/blockatlas/marketdata/rate/fixer"
 	"github.com/trustwallet/blockatlas/pkg/errors"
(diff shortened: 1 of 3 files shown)


storage/market.go
```diff
@@ -59,9 +60,14 @@ func (s *Storage) SaveRates(rates blockatlas.Rates, pl ProviderList) {
 		}
 		err = s.AddHM(EntityRates, rate.Currency, &rate)
 		if err != nil {
+			errCounter++
 			logger.Error(err, "SaveRates", logger.Params{"rate": rate})
 			continue
 		}
+		if errCounter == len(rates) {
+			return errors.E("SaveRates all")
```
Pantani: maybe it's better to try to save one of these errors to put inside the return error context

marketdata/rates.go
```diff
@@ -55,8 +55,12 @@ func runRate(storage storage.Market, p rate.Provider) error {
...
 	if len(rates) > 0 {
-		storage.SaveRates(rates, rateProviders)
+		err = storage.SaveRates(rates, rateProviders)
 		logger.Info("Market rates", logger.Params{"rates": len(rates), "provider": p.GetId()})
 	}
+	if err != nil {
+		logger.Error(err, "runRate", logger.Params{"rates": rates})
+		return err
+	}
 	return nil
```
Pantani:
```suggestion
	if len(rates) == 0 {
		return nil
	}
	err = storage.SaveRates(rates, rateProviders)
	if err != nil {
		return errors.E(err, "runRate", errors.Params{"rates": rates})
	}
	logger.Info("Market rates", logger.Params{"rates": len(rates), "provider": p.GetId()})
	return nil
```

storage/market.go
EnoRage: It is an awesome idea

EnoRage:
Need to update tests and add error context

EnoRage:
More details for other reviewers:
SaveRates() method now returns an error, reasons for it:
1. SaveTicker returns an error, however - it pretty common functions that just operate with different structure type for it, so we need to have a common agreement here (return error everywhere or not return)
2. If Redis will be disconnected and we will not write anything to it - it is impossible to understand it from call function
3. So when we will have an option to catch an error during running worker - we can change cache from Redis to Memory instantly (feature)

Issue #725, opened 2020-01-18
APR of Tron and Cosmos Staking needs to be updated


zachzwei:
## Summary of Bug

Tron Staking APR shows as 0.74% in the app. The current staking reward on Binance.com is 6.7646%
Cosmos Staking shows 11% but no validator has this APR showing. Please double check and correct the percentage.

Staking Rewards on Binance.com
![image](https://user-images.githubusercontent.com/35627271/72660469-80e79a00-3a09-11ea-8806-4ca0cf9e8f7c.png)

Staking Rewards within in the app
![image](https://user-images.githubusercontent.com/35627271/72660479-ad031b00-3a09-11ea-908f-c0e3ae24a88e.png)

![image](https://user-images.githubusercontent.com/35627271/72947043-bb5f8700-3dbb-11ea-8c73-f00851742a75.png)

## Steps to Reproduce

Checked Android 1.7.151, Tron Staking APR is also 0.74%

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

EnoRage:
Currently, Cosmos APR is hardcoded [here](https://github.com/trustwallet/blockatlas/blob/master/platform/cosmos/stake.go#L43)
Tron APR is hardcoded [here](https://github.com/trustwallet/blockatlas/blob/master/platform/tron/api.go#L17) as well.

Pantani:
we need to find a way to have this dynamically. 

The same for Tezos:
https://github.com/trustwallet/blockatlas/tree/master/platform/tezos/api.go#L130

duplicated issue:
https://github.com/trustwallet/blockatlas/issues/455

EnoRage:
The main points, for now, that were discussed with @vikmeup:
- Add cache for Platform (there are multiple suitable methods)
- Use @vikmeup Cosmos pull request for it, do the same for Kava
- TRON, TEZOS and other blockchains will have dynamic APR later, currently, it is not critical
- #455 needs to be modified and contain all blockchains that need  dynamic APR

I will use https://github.com/trustwallet/blockatlas/pull/669 request to:
- [ ] Add dynamic APR for Cosmos and Kava
- [ ] Add caching to Platform service for ( APR, GetDetails(), potentially some other methods)

@Pantani, @vikmeup, please, confirm

vikmeup:
👍

EnoRage:
[...]

PR #747, opened 2020-01-26
[Platform] Add Dynamic APR for ATOM/Kava + Caching

EnoRage:
Solution for issues: #455 and #725

- I have added caching (Storage) with its implementation for Platform
- During first response Platform will cache StakingDetails for allowedCachePeriod = 86400 (1 day) now
- Also small refactoring of cmd package (Storage is not a global variable right now)

I also merged @vikmeup [pull request](https://github.com/trustwallet/blockatlas/pull/669) here.

## Testing
1. Run Redis 
2. `go build -o blockatlas . && ./blockatlas api :8420
3. POST requset `http://localhost:8420//v2/staking/list` with body 
`[
  {
    "address": "string",
    "coin": 118
  }
]`
p.s you can use coin 459 also

The response speed will be significantly faster after 1 request


storage/platform.go
```diff
@@ -0,0 +1,48 @@
...
+)
+
+const (
+	EntityDetails        = "ATLAS_PLATFORM_DETAILS"
+	entityDetailsUpdated = "ATLAS_PLATFORM_DETAILS_UPDATED"
+	allowedCachingPeriod = 86400 // 24 hours to UNIX time
+)
+
+func (s *Storage) SaveCachedDetails(coin string, details *blockatlas.StakingDetails) error {
+	return s.AddHM(EntityDetails, createHashMap(coin, ""), details)
```
Pantani: you can use only the coin name here

```suggestion
	return s.AddHM(EntityDetails, coin, details)
```

https://github.com/trustwallet/blockatlas/blob/master/storage/market.go#L73

storage/platform.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+func (s *Storage) GetCachedDetails(coin string) (*blockatlas.StakingDetails, error) {
+	var dt *blockatlas.StakingDetails
+	if !s.isCacheFresh(coin) {
+		return nil, errors.E("outdated cache")
```
Pantani:
```suggestion
		return nil, errors.E("outdated cache", errors.Params{"coin": coin})
```

storage/platform.go
```diff
@@ -0,0 +1,48 @@
...
+	return dt, nil
+}
+
+func (s *Storage) isCacheFresh(coin string) bool {
+	now := time.Now().Unix()
+	if prev, err := s.getTimestamp(coin); err == nil && now-prev <= allowedCachingPeriod {
+		return true
+	}
+
+	if err := s.AddHM(entityDetailsUpdated, createHashMap(coin, ""), now); err != nil {
```
Pantani:
```suggestion
	if err := s.AddHM(entityDetailsUpdated, coin, now); err != nil {
```

storage/platform.go
```diff
@@ -0,0 +1,48 @@
...
+		errors.E(err)
+	}
+	return false
+}
+
+func (s *Storage) getTimestamp(coin string) (t int64, err error) {
+	if err = s.GetHMValue(entityDetailsUpdated, createHashMap(coin, ""), &t); err != nil {
```
Pantani:
```suggestion
	if err = s.GetHMValue(entityDetailsUpdated, coin, &t); err != nil {
```

storage/platform.go
```diff
@@ -0,0 +1,48 @@
...
+func (s *Storage) SaveCachedDetails(coin string, details *blockatlas.StakingDetails) error {
+	return s.AddHM(EntityDetails, createHashMap(coin, ""), details)
+}
+
+func (s *Storage) GetCachedDetails(coin string) (*blockatlas.StakingDetails, error) {
+	var dt *blockatlas.StakingDetails
+	if !s.isCacheFresh(coin) {
+		return nil, errors.E("outdated cache")
+	}
+	if err := s.GetHMValue(EntityDetails, createHashMap(coin, ""), &dt); err != nil {
```
Pantani:
```suggestion
	if err := s.GetHMValue(EntityDetails, coin, &dt); err != nil {
```

platform/cosmos/stake.go
```diff
@@ -38,13 +42,40 @@ func (p *Platform) GetValidators() (blockatlas.ValidatorPage, error) {
...
+	nsd := blockatlas.StakingDetails{
+		Reward: blockatlas.StakingReward{
+			Annual: p.GetMaxAPR(),
+		},
 		MinimumAmount: blockatlas.Amount("0"),
 		LockTime:      1814400,
 		Type:          blockatlas.DelegationTypeDelegate,
 	}
+
+	if err := p.Cache.SaveCachedDetails(s, &nsd); err != nil {
```
Pantani: Is there a reason to save the entire object inside the database? Can we save only the APR to avoid save useless entries inside the database?

EnoRage: Some other parameters can be dynamic in the future, that's why I decided to save the entire StakingDetails object

Pantani: great, make sense!

storage/platform.go
EnoRage: Nice!

platform/cosmos/stake.go
vikmeup: I would only store what's needed

EnoRage: I suggest to change Platform Storage interface to GetCachedMaxAPR() SetCachedMaxAPR() instead of GetCachedDetails() and SaveCachedDetails() 
@vikmeup

EnoRage:
Suggest to open a new issue for it. Normalize charts Unit test random bahvior
Screen Shot 2020-01-26 at 10 04 03

vikmeup:
Caching layer needs to sit on top of platforms and platforms should not be aware of caching layer.

api/batch.go
```diff
@@ -53,6 +55,7 @@ func makeStakingDelegationsBatchRoute(router gin.IRouter) {
 			if !ok {
 				continue
 			}
+			//
```
vikmeup: Remove?

api/batch.go
```diff
@@ -73,7 +76,7 @@ func makeStakingDelegationsBatchRoute(router gin.IRouter) {
 // @Success 200 {object} blockatlas.DelegationsBatchPage
 // @Router /v2/staking/list [post]
 func makeStakingDelegationsSimpleBatchRoute(router gin.IRouter) {
-	router.POST("/staking/list", func(c *gin.Context) {
+	router.POST("/staking/list", gincache.CacheMiddleware(time.Hour*24*7, func(c *gin.Context) {
```
vikmeup: 👍

platform/cosmos/stake.go
```diff
@@ -38,15 +42,33 @@ func (p *Platform) GetValidators() (blockatlas.ValidatorPage, error) {
...
 		MinimumAmount: blockatlas.Amount("0"),
 		LockTime:      1814400,
 		Type:          blockatlas.DelegationTypeDelegate,
 	}
 }
 
+func (p *Platform) GetMaxAPR() float64 {
+	validators, err := p.GetValidators()
+	if err != nil {
+		return defaultAnnualReward
```
vikmeup: Can we remove default and return 0 instead maybe? 

Ideally not to have any hard coded values

vikmeup: Login behind: if you provide wrong value to the user - it’s deceptive to a user. 0 would be safer in this case - but if it’s cached as 0, than it’s a problem

vikmeup: Maybe throw an error would be better in this case

EnoRage: No. We will need to change GetDetails() interface. If so, we will add Redis and check that err == nil, if we will have to err => use Redis (when we have cached valid value).

However, it is not good to change the basic interface, I cannot say that there is any other option how to handle it without making it more worse than rn.

So, I recommend returning the default apr if err, caching period is not critical enough to be an issue. Also, I cannot say that GetValidators() will return error often enough to care about it

Pantani: I agree to remove the hardcoded value. It's better to show 0 or nothing than show a magic number

EnoRage:
1. Default Value for MaxAPR() is 0 right now
2. No cache handler was added, if we will have 0 that will be caching - it is possible to retry with /nocache like this: 
`/v2/staking/list/nocache`

EnoRage:
@vikmeup @Pantani, please review the updated changes

platform/cosmos/stake.go
vikmeup: `defaultAnnualReward` make it global for all staking platforms

api/batch.go
```diff
@@ -73,28 +75,31 @@ func makeStakingDelegationsBatchRoute(router gin.IRouter) {
...
 // @Router /v2/staking/list [post]
 func makeStakingDelegationsSimpleBatchRoute(router gin.IRouter) {
-	router.POST("/staking/list", func(c *gin.Context) {
-		var reqs CoinsRequest
-		if err := c.BindJSON(&reqs); err != nil {
-			ginutils.ErrorResponse(c).Message(err.Error()).Render()
-			return
-		}
+	router.POST("/staking/list", gincache.CacheMiddleware(time.Hour*24, handleStakingRequest))
+	router.POST("/staking/list/nocache", handleStakingRequest)
```
vikmeup: `staking/list/nocache` please remove - we only need one

platform/cosmos/stake.go
EnoRage: Do we need to use it (defaultAnnualReward) with Algorand, TRON, Tezos?

PR #753, opened 2020-01-27
Add username field to social link struct

vikmeup:


Pantani:
@vikmeup where do we going to use that?

vikmeup:
Clients would need to know the difference between url vs username/handle to present better UI.

Issue #740, opened 2020-01-23
[ONTOLOGY] add block api


Pantani:
## Summary

Implement [block API](https://github.com/trustwallet/blockatlas/blob/master/pkg/blockatlas/api.go#L38) to ontology

## Problem Definition

Block API:
https://github.com/trustwallet/blockatlas/blob/master/pkg/blockatlas/api.go#L38

Example:
https://github.com/trustwallet/blockatlas/tree/master/platform/ethereum/api.go#L178

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #756, opened 2020-01-27
[ONTOLOGY] implement block api for ontology

EnoRage:
Issue  #740
Current problem: Sometimes ont explorer returns incorrect results that are not specified by explorer api
Screen Shot 2020-01-27 at 16 22 57

```
{
    "error": "GetBlockByNumber",
    "meta": {
        "error": {
            "error": "json: cannot unmarshal bool into Go struct field blockByNumberResponse.Result of type struct { PrevBlock string \"json:\\\"PrevBlock\\\"\"; TxnsRoot string \"json:\\\"TxnsRoot\\\"\"; BlockTime int \"json:\\\"BlockTime\\\"\"; NextBlock string \"json:\\\"NextBlock\\\"\"; BookKeeper string \"json:\\\"BookKeeper\\\"\"; TxnNum int \"json:\\\"TxnNum\\\"\"; Height int \"json:\\\"Height\\\"\"; TxnList []struct { TxnTime int \"json:\\\"TxnTime\\\"\"; ConfirmFlag int \"json:\\\"ConfirmFlag\\\"\"; TxnHash string \"json:\\\"TxnHash\\\"\"; Height int \"json:\\\"Height\\\"\" } \"json:\\\"TxnList\\\"\"; Hash string \"json:\\\"Hash\\\"\"; ConsensusData string \"json:\\\"ConsensusData\\\"\"; BlockSize int \"json:\\\"BlockSize\\\"\" }",
            "meta": {
                "method": "GET",
                "url": "https://explorer.ont.io/api/v1/explorer/block/7678552?"
            },
            "type": "Platform Unmarshal Error"
        }
    }
}
```
By current design - it is ok, however, we can try to use not explorer api / something else


EnoRage:
My idea is that I can try to find other endpoints, they have rest for nodes with the same method

EnoRage:
https://github.com/ontio/ontology-go-sdk

platform/ontology/api_test.go
```diff
@@ -176,3 +176,51 @@ func TestNormalize(t *testing.T) {
 		}
 	}
 }
+
+func TestAPI_CurrentBlockNumber(t *testing.T) {
```
Pantani: Don't do integration/functional tests inside the unit test files.

You can create another file and use another build tag, like that:
https://github.com/trustwallet/blockatlas/blob/43b998bedcf11417cdc31601117335b408ceb45b/pkg/integration/integration_test.go#L1

platform/ontology/client.go
```diff
@@ -2,13 +2,59 @@ package ontology
...
+	NextBlock     string `json:"NextBlock"`
+	BookKeeper    string `json:"BookKeeper"`
+	TxnNum        int    `json:"TxnNum"`
+	Height        int    `json:"Height"`
+	Hash          string `json:"Hash"`
+	ConsensusData string `json:"ConsensusData"`
+	BlockSize     int    `json:"BlockSize"`
+}
+
+type blockByNumberResponse struct {
```
Pantani: please, avoid nested structs, It's complex, hard to understand, and not so testable. Try to use only in a simple case

platform/ontology/client.go
```diff
@@ -17,3 +63,50 @@ func (c *Client) GetTxsOfAddress(address, assetName string) (txPage *TxPage, err
...
+		response blockByNumberResponse
+		block    blockatlas.Block
+		txs      []blockatlas.Tx
+	)
+	err := c.Get(&response, url, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	if response.Error == 0 {
```
Pantani: please, avoid parsing inside the client methods. The file should need to be responsible only for client methods, do you need to do that inside the `api.go` file

platform/ontology/client.go
Pantani: and you don't need to parse the entire response object

EnoRage: so, basically there is no difference to parse / unparse the entire object. Also it is your point about nested structs is opposite. It is - not complex, it is easier to understand, it is much more testable. Also I cannot find more simple case than that.

platform/ontology/api_test.go
EnoRage: the main reason I write it here is that integration tests are ignored at pipeline

Pantani:
> https://github.com/ontio/ontology-go-sdk

we always try to avoid adding a new package. Only if do you need something for the core of the application, we never add a new package for only one platform

platform/ontology/api.go
```diff
@@ -55,6 +55,52 @@ func (p *Platform) GetTokenTxsByAddress(address string, token string) (blockatla
...
+	}
+	var (
+		block blockatlas.Block
+		txs   []blockatlas.Tx
+	)
+	if response.Error == 0 {
+		block.ID = response.Result.Hash
+		block.Number = int64(response.Result.Height)
+		for _, txn := range response.Result.TxnList {
+			tx := new(blockatlas.Tx)
```
vikmeup: should part of the code should be moved into `Normalize()` so you can easily test that

Pantani: @EnoRage https://github.com/trustwallet/blockatlas/issues/763
can you handle that, please?

platform/ontology/api.go
```diff
@@ -55,6 +55,52 @@ func (p *Platform) GetTokenTxsByAddress(address string, token string) (blockatla
...
+	if len(block.Result) > 0 {
+		height = (int64)(block.Result[0].Height)
+	}
+	return height, nil
+}
+
+func (p *Platform) GetBlockByNumber(num int64) (*blockatlas.Block, error) {
+	response, err := p.client.GetBlockByNumber(num)
+	if err != nil {
+		logger.Error("GetBlockByNumber", logger.Params{"platform": p.Coin().Symbol, "details": err.Error()})
```
Pantani: @EnoRage also we don't need that log, we can log only the top level

EnoRage:
I will open new pull for it soon

PR #758, opened 2020-01-27
reduce in sorting without free space (using baking-bad API) 

Elemberg:


platform/tezos/bakingbad/api.go
```diff
@@ -0,0 +1,60 @@
...
+)
+
+const (
+	baseURL        = "https://api.baking-bad.org"
+	defaultTimeout = time.Second * 2
+)
+
+type (
+	API struct {
+		client *http.Client
```
Pantani: we already have a package to deal with http requests

https://github.com/trustwallet/blockatlas/blob/master/pkg/blockatlas/client.go

pkg/blockatlas/staking.go
```diff
@@ -26,6 +26,7 @@ type StakingReward struct {
 
 type StakingDetails struct {
 	Reward        StakingReward  `json:"reward"`
+	FreeSpace     float64        `json:"free_space"`
```
Pantani: Can you give me more details? Why do you need to change the global stake model?

Issue #767, opened 2020-01-28
[POLKADOT] Critical fail due to interface conversion 


EnoRage:
<img width="989" alt="Screen Shot 2020-01-28 at 16 14 48" src="https://user-images.githubusercontent.com/22479658/73267203-9dde5300-41e9-11ea-8fcc-48f12e899b67.png">

Steps to reproduce:

- Git clone from master
- Clean Redis
- Run observer

platform/polkadot/api.go:135 looks like bug is here

PR #770, opened 2020-01-28
[Blockatlas] Simplify and refactor project

EnoRage:
What was done:
- remove cobra (no need to use cobra for passing 1 param)
- cmd/syncmarkets|observer|api/main.go (different binaries for each project)
- change readme, docker (to make it work with new changes) 

Reasons for refactoring:
> Cobra is both a library for creating powerful modern CLI applications as well as a program to generate applications and command files

Blockatlas is not CLI application. The only reason to use Cobra here was to pass 1 parameter during launch. It is an obvious overhead. Another reason to remove it is that cmd package is written not in Go lang standards. Look at Kubernetes where there are separate binaries for each tool. Look at these projects as an examples also:
Examples:
- https://github.com/heptio/ark/tree/master/cmd (just a really small main function with everything else in packages)
- https://github.com/moby/moby/tree/master/cmd
- https://github.com/prometheus/prometheus/tree/master/cmd
- https://github.com/influxdata/influxdb/tree/master/cmd
- https://github.com/satellity/satellity/tree/master/cmd/satellity
- https://github.com/dapr/dapr/tree/master/cmd
     
Also we need to apply a strict style guide and review each PR according to it. There is no way how an open source project with more than 2 months history cannot have style guide. Uber style guide is suitable for blockatlas. However any other public style guide can be suitable too

I want also to give a link that have a collection of the best practices in golang. 
https://github.com/golang-standards/project-layout


EnoRage:
Currently it is not fully ready for merging. 
Check list:
- [ ] Check all the imports are correct after renaming some packages (go mod)
- [ ] Run all Unit tests
- [ ] Check all the api's
- [ ] Get the review from the leading maintainers and approval on it
- [ ] Fix Docker and docker-compose, k8s and all CI/CD pipelines

EnoRage:
The main reason to do it now:
1. An open source project must contain the best practices to allow any developer easily add features and fix bugs. Current solution will help to simplify codebase and reduce the entry barrier for launching and working with Blockatlas
2. Good architecture and style guide will help to bring more features faster. It will also prevent contributors to create a new bugs
3. Current changes are not too hard to understand and check. It is written in 1 hour and can be tested and merged in 1 day. Good codebase will help to work over project. So there is no any other critical bug / critical feature to implement for these 1-2 days

cmd/api/main.go
```diff
@@ -0,0 +1,127 @@
...
+	gin.SetMode(viper.GetString("gin.mode"))
+	engine := gin.New()
+
+	engine.Use(gin.Recovery())
+	engine.Use(ginutils.CheckReverseProxy, sg)
+	engine.Use(ginutils.CORSMiddleware())
+
+	engine.OPTIONS("/*path", ginutils.CORSMiddleware())
+
+	engine.GET("swagger/*any", ginSwagger.WrapHandler(swaggerFiles.Handler))
```
Pantani: Did you move the API routes inside the CMD package 😱? CMD needs to be in charge only about the initialization

cmd/api/main.go
```diff
@@ -0,0 +1,127 @@
+package main
```
Pantani: the main package needs to be in the root folder

cmd/api/main.go
```diff
@@ -0,0 +1,127 @@
...
+	platform.Init()
+
+	host := viper.GetString("storage.redis")
+	err = cache.Init(host)
+	if err != nil {
+		logger.Fatal(err)
+	}
+}
+
+func main() {
```
Pantani: a method with almost 100 lines

cmd/api/main.go
EnoRage: Examples of industry leading projects with the same design pattern

https://github.com/heptio/ark/tree/master/cmd (just a really small main function with everything else in packages)
https://github.com/moby/moby/tree/master/cmd
https://github.com/prometheus/prometheus/tree/master/cmd
https://github.com/influxdata/influxdb/tree/master/cmd
https://github.com/satellity/satellity/tree/master/cmd/satellity
https://github.com/dapr/dapr/tree/master/cmd

cmd/api/main.go
EnoRage: So what?

cmd/api/main.go
EnoRage: CMD package is not exists right now. CMD is folder to run multiple services. Each of them is a single binary. I have already sent examples about using it like these. Here we init the routes during starting the service

cmd/api/main.go
EnoRage: Here are the example of large method in current architecture: https://github.com/trustwallet/blockatlas/blob/master/cmd/observer.go#L21

Pantani: yes, this is a good point to refactor and have better clean code, not the cmd package

cmd/api/main.go
Pantani: https://github.com/golang-standards/project-layout#cmd

Pantani:
```
/cmd
Main applications for this project.

The directory name for each application should match the name of the executable you want to have (e.g., /cmd/myapp).

It's common to have a small main function that imports and invokes the code from the /internal and /pkg directories and nothing else.
```

cmd/api/main.go
EnoRage: Current architecture is not using any good design pattern. If you could provide examples of projects that are using the same principles as Blockatlas - it will be great. We provide several examples of industry leading projects that have the same design pattern as we use in this PR. We have already explained that it will reduce the entry barrier for new developers as well as help to maintain project. Using complex framework for small issue is not a advantage of current solution as well

cmd/api/main.go
EnoRage: So, do you see any opposite of it in this file?

Pantani: there are API routes inside this file

terorie:
Please leave the import path as is. You can use Go module replace directives to maintain a local branch. (More at https://github.com/golang/go/wiki/Modules)

cmd/api/main.go
EnoRage: This is a good point to think about redesigning blocklatlas/api package. It is not  intuitive with multiple files with weird naming for some of them as will as almost random functions and methods for api

EnoRage:
Polkadot was unloaded, due to unstable behavior, there is a suitable issue for it #767

platform/registry.go
```diff
@@ -80,7 +79,7 @@ var platformList = []blockatlas.Platform{
 	&algorand.Platform{},
 	&nano.Platform{},
 	&harmony.Platform{},
-	&polkadot.Platform{CoinIndex: coin.KSM},
+	//&polkadot.Platform{CoinIndex: coin.KSM},
```
vikmeup: Can you enable polkadot? but disable in integration tests?

vikmeup:
To fetch tx details:

1. Explorer:  `https://explorer.ont.io/v2/transactions/0767a98f8ea30557c49c35287c871a6566d534aeda582f7a091d9de359ec877b`
2. Node `https://dappnode1.ont.io:10334/api/v1/transaction/0767a98f8ea30557c49c35287c871a6566d534aeda582f7a091d9de359ec877b?raw=0`

cmd/api/main.go
Pantani: can you put this API routes inside another method or package, please?

platform/registry.go
EnoRage: Will be enabled by @Pantani PR when will be fixed

PR #775, opened 2020-01-29
fix polkadot crash

Pantani:
closes https://github.com/trustwallet/blockatlas/issues/767

Polkadot is crashing for two reasons because doesn't have the second object inside the slice data and sometimes can't convert the value two floats. Can you check if is this right @hewigovens ?

### DIFF
--- a/platform/polkadot/api.go
+++ b/platform/polkadot/api.go
@@ -125,6 +125,13 @@ func (p *Platform) NormalizeExtrinsic(srcTx *Extrinsic) *blockatlas.Tx {
 		Sequence: srcTx.Nonce,
 	}
 
+	value := "0"
+	if len(datas) > 0 {
+		vf, ok := datas[0].Value.(float64)
+		if ok {
+			value = fmt.Sprintf("%.0f", vf)
+		}
+	}
 	decimals := p.Coin().Decimals
 	if srcTx.CallModule == ModuleBalances &&
 		srcTx.CallModuleFunction == ModuleFunctionTransfer {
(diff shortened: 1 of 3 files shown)


platform/polkadot/client.go
```diff
@@ -12,7 +12,7 @@ type Client struct {
 
 func (c *Client) GetTransfersOfAddress(address string) ([]Transfer, error) {
 	var res SubscanResponse
-	err := c.Post(&res, "/scan/transfers", TransfersRequest{Address: address, Row: blockatlas.TxPerPage})
+	err := c.Post(&res, "scan/transfers", TransfersRequest{Address: address, Row: blockatlas.TxPerPage})
```
hewigovens: I think we should remove / in `GetBase`? what if there is a trailing / in `config.yml`

Pantani: the method applies the `/`, we don't need them

platform/polkadot/api.go
```diff
@@ -125,14 +125,21 @@ func (p *Platform) NormalizeExtrinsic(srcTx *Extrinsic) *blockatlas.Tx {
...
 	}
 
+	value := "0"
+	if len(datas) > 0 {
+		vf, ok := datas[0].Value.(float64)
+		if ok {
+			value = fmt.Sprintf("%.0f", vf)
+		}
+	}
 	decimals := p.Coin().Decimals
```
vikmeup: add test?

platform/polkadot/api.go
```diff
@@ -125,14 +125,21 @@ func (p *Platform) NormalizeExtrinsic(srcTx *Extrinsic) *blockatlas.Tx {
 		Sequence: srcTx.Nonce,
 	}
 
+	value := "0"
+	if len(datas) > 0 {
```
vikmeup: would it make sense to have similar function as: https://github.com/trustwallet/blockatlas/pull/777/files#diff-f77bff57c22cf320c2a91afaaef92572R22

Pantani: the logic's a little different

platform/polkadot/api.go
Pantani: done!

Issue #760, opened 2020-01-27
[Domains] Add support to FIO protocol


Pantani:
## Summary

Add support to FIO protocol

## Problem Definition / Why?

We need to integrate domains lookup for FIO protocol

## Where/how shows?

When we send a transaction to another user

## Proposal

Implement a naming service API
https://github.com/trustwallet/blockatlas/blob/master/pkg/blockatlas/api.go#L75

https://github.com/trustwallet/blockatlas/blob/master/api/naming_service.go

catenocrypt:
Solved with PR #780 .

PR #780, opened 2020-01-30
Add FIO naming support (testnet)

catenocrypt:
Extending naming providers with FIO naming, FIO testnet for now.  Issue #760 .
Format:  'name@fiotestnet'; API endpoint: http://testnet.fioprotocol.io/v1/chain

For testing:
make functional

```
curl http://localhost:8420/v2/ns/lookup?name=adam@fiotestnet\&coins=60
[{"result":"0xce5cB6c92Da37bbBa91Bd40D4C9D4D724A3a8F51","coin":60}]
```

### DIFF
--- a/api/naming_service.go
+++ b/api/naming_service.go
@@ -56,6 +56,7 @@ func MakeLookupRoute(router gin.IRouter) {
 	TLDMapping[".zil"] = CoinType.ZIL
 	// it's on ethereum but same unstoppable api
 	TLDMapping[".crypto"] = CoinType.ZIL
+	TLDMapping["@fiotestnet"] = CoinType.FIO
 }
 
 // @Summary Lookup .eth / .zil addresses
@@ -105,15 +106,15 @@ func sliceAtoi(sa []string) ([]uint64, error) {
 }
 
 func handleLookup(name string, coins []uint64) (result []blockatlas.Resolved, err error) {
+	// Assumption: format of the name can be decided (top-level-domain), and at most one naming service is tried
 	name = strings.ToLower(name)
-	ss := strings.Split(name, ".")
-	if len(ss) == 0 {
-		return nil, errors.E("name not found", errors.Params{"name": name, "coins": coins})
+	tld, ok := getTLD(name)
+	if !ok {
+		return nil, errors.E("name format not recognized", errors.Params{"name": name, "coins": coins})
 	}
-	tld := "." + ss[len(ss)-1]
 	id, ok := TLDMapping[tld]
 	if !ok {
-		return nil, errors.E("name not found", errors.Params{"name": name, "coins": coins})
+		return nil, errors.E("name not found", errors.Params{"name": name, "coins": coins, "tld": tld})
 	}
 	api, ok := platform.NamingAPIs[id]
 	if !ok {
@@ -121,4 +122,22 @@ func handleLookup(name string, coins []uint64) (result []blockatlas.Resolved, er
 	}
 	result, err = api.Lookup(coins, name)
 	return
+}
+
+// Obtain tld from then name, e.g. ".ens" from "nick.ens"
+func getTLD(name string) (tld string, ok bool) {
+	if strings.Index(name, ".") >= 0 {
+		// xxx.tld format
+		ss := strings.Split(name, ".")
+		tld := "." + ss[len(ss)-1]
+		return tld, true
+	}
+	if strings.Index(name, "@") >= 0 {
+		// xxx@tld format
+		ss := strings.Split(name, "@")
+		tld := "@" + ss[len(ss)-1]
+		return tld, true
+	}
+	// none matched
+	return "", false
 }
--- a/api/naming_service.go
+++ b/api/naming_service.go
@@ -3,6 +3,7 @@ package api
 import (
 	"github.com/trustwallet/blockatlas/pkg/errors"
 	"net/http"
+	"math"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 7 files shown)


api/naming_service.go
```diff
@@ -122,3 +123,21 @@ func handleLookup(name string, coins []uint64) (result []blockatlas.Resolved, er
 	result, err = api.Lookup(coins, name)
 	return
 }
+
+// Obtain tld from then name, e.g. ".ens" from "nick.ens"
+func getTLD(name string) (tld string, ok bool) {
```
vikmeup: Test?

platform/fio/api.go
```diff
@@ -22,3 +22,20 @@ func (p *Platform) Coin() coin.Coin {
...
+		if err != nil {
+			return result, err
+		}
+		result = append(result, blockatlas.Resolved{Coin: coinId, Result: address})
+	}
+
+	return result, nil
+}
+
+
```
vikmeup: Did you run formatting and linting?

platform/fio/client.go
```diff
@@ -2,8 +2,32 @@ package fio
...
 import (
 	"github.com/trustwallet/blockatlas/pkg/blockatlas"
+	"github.com/trustwallet/blockatlas/pkg/errors"
 )
 
 type Client struct {
 	blockatlas.Request
 }
+
+type GetPubAddressRequest struct {
```
EnoRage: Maybe it is ok to add structures to the model.go file like any other platform?
[example](https://github.com/trustwallet/blockatlas/blob/master/platform/cosmos/model.go)

platform/fio/api.go
```diff
@@ -22,3 +22,20 @@ func (p *Platform) Coin() coin.Coin {
 func (p *Platform) GetTxsByAddress(address string) (page blockatlas.TxPage, err error) {
 	return page, err
 }
+
+func (p *Platform) Lookup(coins []uint64, name string) ([]blockatlas.Resolved, error) {
+	var result []blockatlas.Resolved
+	for _, coinId := range coins {
+		coinObj := coin.Coins[uint(coinId)]
+		address, err := p.client.lookupPubAddress(name, coinObj.Symbol)
+		if err != nil {
```
EnoRage: Maybe we can ignore coins with errors and return the correct ones
If I will have an array of coins where only 1 is incorrect - it is better to return all the info except the invalid coin?

api/naming_service.go
```diff
@@ -122,3 +123,21 @@ func handleLookup(name string, coins []uint64) (result []blockatlas.Resolved, er
...
+	if strings.Index(name, ".") >= 0 {
+		// xxx.tld format
+		ss := strings.Split(name, ".")
+		tld := "." + ss[len(ss)-1]
+		return tld, true
+	}
+	if strings.Index(name, "@") >= 0 {
+		// xxx@tld format
```
hewigovens: Is it possbile to have something like: xxx.ens@fio?

platform/fio/client.go
```diff
@@ -2,8 +2,32 @@ package fio
...
+
+type GetPubAddressResponse struct {
+	PublicAddress 	string  `json:"public_address"`
+	BlockNum  		int 	`json:"block_num"`
+	Message         string  `json:"message"`
+}
+
+func (c *Client) lookupPubAddress(name string, coinSymbol string) (address string, error error) {
+	var res GetPubAddressResponse
+	err := c.Post(&res, "get_pub_address", GetPubAddressRequest{FioAddress: name, TokenCode: coinSymbol})
```
hewigovens: does the api endpoint support batching?

api/naming_service.go
catenocrypt: unit test added

platform/fio/api.go
catenocrypt: formatting done

platform/fio/client.go
catenocrypt: Done, thanks

api/naming_service.go
catenocrypt: TLD separation logic reworked, this special case is handled now, unit test added.

platform/fio/client.go
catenocrypt: I don't think so.  Not a typical use case.

platform/fio/api.go
catenocrypt: Good suggestion.  This is a matter of taste/usage pattern.  After some consideration I think in partial error case the caller is more likely to handle error if error is returned than if partial error is returned.  That has the risk of not noting error, and attempting to use missing address.  Better to return error if not all could be resolved.  Typical use case is with one coin anyways.

platform/fio/client.go
```diff
@@ -2,8 +2,21 @@ package fio
...
 
 type Client struct {
 	blockatlas.Request
 }
+
+func (c *Client) lookupPubAddress(name string, coinSymbol string) (address string, error error) {
+	var res GetPubAddressResponse
+	err := c.Post(&res, "get_pub_address", GetPubAddressRequest{FioAddress: name, TokenCode: coinSymbol})
+	if err != nil {
+		return "", err
```
Pantani:
```suggestion
		return "", errors.E(err, "Error lokking up FIO name", errors.Params{"name": name, "coinSymbol": coinSymbol})
```

platform/fio/client.go
```diff
@@ -2,8 +2,21 @@ package fio
...
+	}
+	if res.Message != "" {
+		return "", errors.E("Error lokking up FIO name: "+res.Message, errors.Params{"name": name, "coinSymbol": coinSymbol})
```
Pantani:
```suggestion
		return "", errors.E("Error lokking up FIO name", errors.Params{"name": name, "coinSymbol": coinSymbol, "Message": res.Message})
```

api/naming_service.go
```diff
@@ -122,3 +124,17 @@ func handleLookup(name string, coins []uint64) (result []blockatlas.Resolved, er
 	result, err = api.Lookup(coins, name)
 	return
 }
+
+// Obtain tld from then name, e.g. ".ens" from "nick.ens"
+func getTLD(name string) (tld string, ok bool) {
```
dpereskokov: may be better to return `err` instead of `ok`?

PR #806, opened 2020-02-05
Checksum coingeico Ethreum address

kolya182:
Fix inconsistency issue when ticker from CG returns in lowercase, better for client to convert checksum to lc if needed, rather lowercase to checksum
Adds check if token supplied.
Simplify code
Ignore binary files after project builded

### DIFF
--- a/syncmarkets/clients/coingecko/cache.go
+++ b/syncmarkets/clients/coingecko/cache.go
@@ -94,6 +94,9 @@ func getCoinsMap(coins GeckoCoins) map[string]GeckoCoin {
 }
 
 func normalizeTokenId(platform, addr string) string {
+	if platform == "" || addr == "" {
+		return ""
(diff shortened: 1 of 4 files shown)


syncmarkets/clients/coingecko/cache.go
```diff
@@ -92,3 +92,12 @@ func getCoinsMap(coins GeckoCoins) map[string]GeckoCoin {
 	}
 	return coinsMap
 }
+
+func normalizeTokenId(platform, addr string) string {
```
Pantani: can you create unit tests for that?

kolya182: https://github.com/trustwallet/blockatlas/pull/806/commits/861dd8f8a4e501d120e3f45395623930b88a6e63

PR #808, opened 2020-02-05
Add integration tests for naming service

Pantani:
closes https://github.com/trustwallet/blockatlas/issues/541

# Headline 

Add integrations tests for naming service domains

## Changes
- Move ens methods to `domains` package inside the `services` folder
- Create a package for `naming service` integrations tests

### DIFF
--- a/pkg/tests/integration/integration_test.go
+++ b/pkg/tests/integration/integration_test.go
@@ -4,8 +4,8 @@ package integration
 
 import (
 	"github.com/trustwallet/blockatlas/config"
-	"github.com/trustwallet/blockatlas/pkg/tests/integration/bitcoin"
-	"github.com/trustwallet/blockatlas/pkg/tests/integration/ontology"
+	"github.com/trustwallet/blockatlas/pkg/tests/integration/domains"
+	"github.com/trustwallet/blockatlas/platform"
 	"os"
 	"testing"
 )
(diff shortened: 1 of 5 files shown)


pkg/tests/integration/integration_test.go
```diff
@@ -17,6 +17,9 @@ func Test(t *testing.T) {
 	} else {
 		config.LoadConfig(configPath)
 	}
-	ontology.TestOntology(t)
-	bitcoin.TestBitcoin(t)
+	platform.Init()
+
+	//ontology.TestOntology(t)
+	//bitcoin.TestBitcoin(t)
```
vikmeup: any reason to disable?

services/domains/domains.go
```diff
@@ -0,0 +1,56 @@
+package domains
+
+import (
+	CoinType "github.com/trustwallet/blockatlas/coin"
```
vikmeup: did you consider rename coin to cointype ? to be `CoinType`

pkg/tests/integration/integration_test.go
Pantani: sorry I did forget, but it's already enabled now

services/domains/domains.go
Pantani: no, this name comes from the current implementation, but I prefer to call this package `coins`, in the plural

Issue #812, opened 2020-02-06
Cache middleware doesn't work


dpereskokov:
all requests go without caching

Pantani:
did you test without that? We need to restore the from a byte array
https://github.com/trustwallet/blockatlas/pull/813/files#r376481755

Issue #739, opened 2020-01-23
Charts api for all time is slow


dpereskokov:
Two potential problems: 
1) no cache 
2) slow provider or our processing

PR #813, opened 2020-02-06
Charts api for all time are slow and solved basic caching logic

dpereskokov:
Fixes: #812, #739

### DIFF
--- a/pkg/ginutils/gincache/cache.go
+++ b/pkg/ginutils/gincache/cache.go
@@ -10,7 +10,6 @@ import (
 	"github.com/patrickmn/go-cache"
 	"github.com/trustwallet/blockatlas/pkg/errors"
 	"github.com/trustwallet/blockatlas/pkg/logger"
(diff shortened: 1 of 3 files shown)


pkg/ginutils/gincache/cache.go
```diff
@@ -120,16 +119,12 @@ func (mc *memCache) getCache(key string) (*cacheResponse, error) {
 	if !ok {
 		return nil, fmt.Errorf("gin-cache: invalid cache key %s", key)
 	}
-	r, ok := c.([]byte)
+	r, ok := c.(cacheResponse)
```
Pantani: the cache is saved in a byte array

https://github.com/trustwallet/blockatlas/pull/813/files#diff-70e12b2752974d408c475deba3e9a7a5R107

dpereskokov: Try to debug this line. Cast always fails since `c` comes as `cacheResponse` struct

PR #817, opened 2020-02-07
Refactor STEP 2

EnoRage:
STEP 2:
- split REST api to platform api, markets api, observer api
- observer and platform api can work as an instance per container (each coin - new container)


platform/registry.go
```diff
@@ -102,108 +102,116 @@ var (
...
-		platformList = append(platformList, &polkadot.Platform{CoinIndex: coin.KSM})
+		return &polkadot.Platform{CoinIndex: coin.KSM}
 	default:
-		platformList = allPlatformsList
+		return nil
 	}
+}
 
+func getActivePlatforms(platformHandle string) []blockatlas.Platform {
+	var platformList []blockatlas.Platform
```
vikmeup:
```suggestion

	logger.Info("Loaded with: ", logger.Params{"handle": platformHandle})
	
	platform := getPlatform(platformHandle)
	if platform != nil {
		return []{platform}
	} else {
		return allPlatformsList
	}
```

Did you consider simpler version?

internal/init.go
```diff
@@ -56,8 +55,8 @@ func InitAPIWithRedis(defaultPort, defaultConfigPath string) (string, string, *g
 	config.LoadConfig(confPath)
 	logger.InitLogger()
 
-	host := viper.GetString("storage.redis")
-	err = cache.Init(host)
+	//host := viper.GetString("storage.redis")
+	err = cache.Init("redis://localhost:6379")
```
vikmeup: avoid committing this

EnoRage: Sorry, my fuckup, solved

platform/registry.go
```diff
@@ -34,74 +33,187 @@ import (
...
+	if platform != nil {
+		return []blockatlas.Platform{platform}
+	}
+	return allPlatformsList
+}
 
-// CollectionAPIs contain platforms which collections services
-var CollectionAPIs map[uint]blockatlas.CollectionAPI
+func Init(platformHandle string) {
+	platformList := getActivePlatforms(platformHandle)
```
Pantani: is there a way to initialization only 2 or 3 platforms in one application? like Ethereum one, Bitcoin another one and some others small coins another one

EnoRage:
There is no any point to run them in single binary except legacy support (our case). I cannot see any significant benefits of that approach for a long term. 
We will add docker-compose where you will have an option to instantly deploy ETH and BTC, or all, or any other combination. There will be a feature issue for that

Issue #825, opened 2020-02-09
SIGSEGV when running `make functional`


arein:
#174  Summary of Bug

1. Clone repository
2. Run `make go-build`
3. Run `make functional`

Using the `config.yml` provided by the repo. Functional tests succeed on Azure pipelines so maybe this is the difference.

Running on OS X / Go 1.13

```
time="2020-02-09T12:19:05-08:00" level=error msg="{\"error\":\"Invalid response\",\"meta\":{\"query\":\"\",\"response\":{},\"route\":\"/v1/tezos/tz1foWxaV3VQyWqFbWTERS6YDJjPT6C7jPp8\"}}" query= response="&{{http://localhost:8420 {} 0xc000496510 0xc00052ce00 0xc0004f0a28 []} {0xc0004f0a28 true} <nil> [] [] <nil> <nil>}" route=/v1/tezos/tz1foWxaV3VQyWqFbWTERS6YDJjPT6C7jPp8
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x521b494]

goroutine 531 [running]:
github.com/trustwallet/blockatlas/pkg/tests/functional.(*Client).testGet(0xc00000f500, 0xc0002a12c0, 0x2e, 0x555e408, 0x0)
	/Users/alexrein/code/blockatlas/pkg/tests/functional/http.go:57 +0x174
github.com/trustwallet/blockatlas/pkg/tests/functional.(*Client).doTests(0xc00000f500, 0x555eacc, 0x3, 0xc00047a6e0, 0x12, 0xc000382ba0)
	/Users/alexrein/code/blockatlas/pkg/tests/functional/http.go:95 +0x39f
created by github.com/trustwallet/blockatlas/pkg/tests/functional.TestApis.func4
	/Users/alexrein/code/blockatlas/pkg/tests/functional/functional_test.go:103 +0xd7
FAIL	github.com/trustwallet/blockatlas/pkg/tests/functional	34.211s
FAIL
make: *** [go-functional] Error 1
```

## Expected Behavior

Functional tests run successfully.

EnoRage:
Can we close it right now?

EnoRage:
@arein

arein:
Still doesn't work for me on _master_.

PR #826, opened 2020-02-09
Fixes HTTP error codes and errors being swallowed

arein:
The content of this PR is threefold:

1. Currently when _/v1/market/info_ fails because `assets.GetCoinInfo` fails we show success to the user not failure. This appears to be the wrong behavior but may not be. @vikmeup please advise
1. When providing wrong user input e.g. coins that are not modeled we return HTTP 500 to the user signaling a (retryable) server error. Instead now we send a HTTP 400 signaling user error
1. We don't log some of the server errors making it hard for us to get the transparency we need to diagnose

## Testing

1. `make go-build && make test`
1. was not able to run functional tests see #825

### DIFF
--- a/api/maketdata.go
+++ b/api/maketdata.go
@@ -61,25 +61,27 @@ func getTickerHandler(storage storage.Market) func(c *gin.Context) {
 		coinQuery := c.Query("coin")
 		coinId, err := strconv.Atoi(coinQuery)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


vikmeup:
@arein can you rebase against master and let's get this merged! PR Looks great!

arein:
@vikmeup thx! :-)

Issue #822, opened 2020-02-09
cBat price incorrect


dpereskokov:
returns $0.02 when price is $0.4

PR #828, opened 2020-02-10
Compound fix prices

dpereskokov:
Compound returs price in ETH. There is no option to fetch them in fiat like USD.
This PR makes market prices API to respect stored price currency in QUOTE.

Fixes: #822

### DIFF
--- a/api/maketdata.go
+++ b/api/maketdata.go
@@ -11,6 +11,7 @@ import (
 	"github.com/trustwallet/blockatlas/services/assets"
 	"github.com/trustwallet/blockatlas/storage"
 	"github.com/trustwallet/blockatlas/syncmarkets"
+	"math/big"
 	"net/http"
 	"strconv"
 	"strings"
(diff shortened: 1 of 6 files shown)


api/maketdata.go
```diff
@@ -123,7 +127,21 @@ func getTickersHandler(storage storage.Market) func(c *gin.Context) {
...
+					exchangeRate *= newRate.Rate
+					percentChange = newRate.PercentChange24h
+				} else {
+					tickerRate, err := storage.GetTicker(strings.ToUpper(r.Price.Currency), "")
+					if err == nil {
+						exchangeRate *= tickerRate.Price.Value
+						percentChange = big.NewFloat(tickerRate.Price.Change24h)
+					}
+				}
+			}
```
arein: this conversion logic is quite a bit of complexity for the controller. How do you feel about moving this in some conversion utility?

PR #832, opened 2020-02-11
Fix Nimiq transaction requests

terorie:
A misinterpretation of the Nimiq RPC API can cause BlockAtlas to return unordered transaction lists or less than 25 transactions even when more are available.
This PR fixes the `getTransactionByAddress` call and ensures transaction count & ordering.

 - [x] tests pass
 - [x] project runs
 - [x] response valid

### DIFF
--- a/platform/nimiq/tests/getTransactionsByAddress_50.json
+++ b/platform/nimiq/tests/getTransactionsByAddress_50.json
@@ -1,0 +1,62 @@
+[
+  {
+    "hash": "61454a43916e075f0e0a830888db7a37e9ffc42503fa3cd2d9eb85fe8067e435",
+    "blockHash": "ec2c6712a977d62e95e0d43eca02bf4161c86197e576073429c0cdc6b9cc1ce1",
+    "blockNumber": 507486,
+    "timestamp": 1554283830,
+    "confirmations": 450839,
+    "from": "99643a627e867dae6cf10ca73b4aff0345d3b72e",
+    "fromAddress": "NQ76 K5J3 LQKX GRXS UT7H 1JKK NJPY 0D2V 7DRE",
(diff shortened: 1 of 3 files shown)


platform/nimiq/tests/getTransactionsByAddress_50.json
```diff
@@ -0,0 +1,422 @@
...
+    "fromAddress": "NQ02 YP68 BA76 0KR3 QY9C SF0K LP8Q THB6 LTKU",
+    "to": "99643a627e867dae6cf10ca73b4aff0345d3b72e",
+    "toAddress": "NQ76 K5J3 LQKX GRXS UT7H 1JKK NJPY 0D2V 7DRE",
+    "value": 100000,
+    "fee": 138,
+    "data": null,
+    "flags": 0
+  },
+  {
+    "hash": "e1041f23c26acb4ddb6918d165a8df9aee16ee25496d60990189416d0c6f7e50",
```
vikmeup: Include only a few transactions, thanks

vikmeup:
@terorie can you finish changes and let's get this merged?

platform/nimiq/tests/getTransactionsByAddress_50.json
```diff
@@ -0,0 +1,422 @@
+[
```
Pantani: why do you use a file? can you use only the necessary cases and put inside the test file, please?

vikmeup: Files are independent entities, we should keep it this well for code organization. 

Code should not be mixed with JSON data

platform/nimiq/tests/getTransactionsByAddress_50.json
terorie: done, only including four now

PR #833, opened 2020-02-11
Improves market error logging

arein:
This slightly improves diagnosing issues when syncing markets.

### DIFF
--- a/syncmarkets/market.go
+++ b/syncmarkets/market.go
@@ -53,13 +53,15 @@ func runMarket(storage storage.Market, p market.Provider) error {
 	if err != nil {
 		return errors.E(err, "GetData")
 	}
+	var saveErrs = 0
 	for _, result := range data {
 		err = storage.SaveTicker(result, marketProviders)
 		if err != nil {
+			saveErrs++
 			logger.Error(errors.E(err, "SaveTicker",
 				errors.Params{"result": result}))
 		}
 	}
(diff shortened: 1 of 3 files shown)


syncmarkets/market.go
```diff
@@ -53,13 +53,15 @@ func runMarket(storage storage.Market, p market.Provider) error {
...
 	for _, result := range data {
 		err = storage.SaveTicker(result, marketProviders)
 		if err != nil {
+			saveErrs++
 			logger.Error(errors.E(err, "SaveTicker",
 				errors.Params{"result": result}))
 		}
 	}
-	logger.Info("Market data result", logger.Params{"markets": len(data), "provider": p.GetId()})
+	logger.Info("Market data result", logger.Params{"markets": len(data), "provider": p.GetId(), "failed": saveErrs})
```
vikmeup:
```suggestion
	logger.Info("Market data result", logger.Params{"markets": len(data), "provider": p.GetId(), "attempt": saveErrs})
``` 

Or something similar ? or `failed_attempt`, unless there is better words to describe it?

arein: I thought about it for a bit: currently the error returned by `err = storage.SaveTicker(result, marketProviders)` may really mean that there was a save error or that (most likely) the ticker is outdated.

I first used `saveErr` to depict this but then realized that this hints at an IO issue rather than the data being outdated. I think `failed` is a good depiction here.

Issue #837, opened 2020-02-12
Remove exposing any internal data or endpoints in the error response


vikmeup:
## Summary of Bug

Error response should not be exposing any internal endpoints or data in the error response

```
{
   "code":500,
   "error":{
      "error":"Get https://mainnet.tezos.org.ua/chains/main/blocks/head/votes/current_period_kind?: net/http: request canceled (Client.Timeout exceeded while awaiting headers): Unable to fetch delegations list",
      "meta":{
         "method":"GET",
         "url":"https://mainnet.tezos.org.ua/chains/main/blocks/head/votes/current_period_kind?"
      }
   }
}
```

PR #844, opened 2020-02-12
Remove sensitive info

Pantani:
closes https://github.com/trustwallet/blockatlas/issues/837

### DIFF
--- a/pkg/blockatlas/client.go
+++ b/pkg/blockatlas/client.go
@@ -48,7 +48,7 @@ func (r *Request) Get(result interface{}, path string, query url.Values) error {
 		queryStr = query.Encode()
 	}
 	uri := strings.Join([]string{r.GetBase(path), queryStr}, "?")
-	return r.Execute("GET", uri, nil, result)
+	return r.Execute("GET", uri, nil, result, errors.Params{"path": path})
 }
 
 func (r *Request) Post(result interface{}, path string, body interface{}) error {
@@ -57,14 +57,15 @@ func (r *Request) Post(result interface{}, path string, body interface{}) error 
 		return err
 	}
 	uri := r.GetBase(path)
-	return r.Execute("POST", uri, buf, result)
+	return r.Execute("POST", uri, buf, result, errors.Params{"path": path})
 }
 
-func (r *Request) Execute(method string, url string, body io.Reader, result interface{}) error {
+func (r *Request) Execute(method string, url string, body io.Reader, result interface{}, params errors.Params) error {
+	params["method"] = method
 	start := time.Now()
 	req, err := http.NewRequest(method, url, body)
 	if err != nil {
-		return errors.E(err, errors.TypePlatformRequest, errors.Params{"url": url, "method": method})
+		return errors.E(err, errors.TypePlatformRequest, params)
 	}
 
 	for key, value := range r.Headers {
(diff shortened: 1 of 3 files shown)


pkg/storage/redis/redis.go
```diff
@@ -39,7 +39,7 @@ func (db *Redis) GetValue(key string, value interface{}) error {
 func (db *Redis) Add(key string, value interface{}) error {
 	j, err := json.Marshal(value)
 	if err != nil {
-		return errors.E(err, errors.Params{"value": value})
+		return errors.E(err, errors.Params{"key": key})
```
arein: Can we log the value here so that at least internally we know what failed?

pkg/storage/redis/hmap.go
```diff
@@ -33,7 +33,7 @@ func (db *Redis) GetHMValue(entity, key string, value interface{}) error {
 func (db *Redis) AddHM(entity, key string, value interface{}) error {
 	j, err := json.Marshal(value)
 	if err != nil {
-		return errors.E(err, errors.Params{"value": value})
+		return errors.E(err, errors.Params{"key": key})
 	}
```
arein: Can we log the value here so that at least internally we know what failed?

pkg/blockatlas/client.go
```diff
@@ -73,22 +74,22 @@ func (r *Request) Execute(method string, url string, body io.Reader, result inte
...
 	defer res.Body.Close()
 	b, err := ioutil.ReadAll(res.Body)
 	if err != nil {
-		return errors.E(err, errors.TypePlatformUnmarshal, errors.Params{"url": url, "method": method})
+		return errors.E(err, errors.TypePlatformUnmarshal, params)
 	}
 	err = json.Unmarshal(b, result)
 	if err != nil {
-		return errors.E(err, errors.TypePlatformUnmarshal, errors.Params{"url": url, "method": method})
+		return errors.E(err, errors.TypePlatformUnmarshal, params)
```
arein: are these errors logged somewhere?

pkg/blockatlas/client.go
```diff
@@ -57,14 +57,15 @@ func (r *Request) Post(result interface{}, path string, body interface{}) error
...
 }
 
-func (r *Request) Execute(method string, url string, body io.Reader, result interface{}) error {
+func (r *Request) Execute(method string, url string, body io.Reader, result interface{}, params errors.Params) error {
+	params["method"] = method
 	start := time.Now()
 	req, err := http.NewRequest(method, url, body)
 	if err != nil {
-		return errors.E(err, errors.TypePlatformRequest, errors.Params{"url": url, "method": method})
+		return errors.E(err, errors.TypePlatformRequest, params)
```
arein: how does this actually address the leaking?
Looking at #837 it seems like the path might be sensitive as well

Pantani: the path is already inside the code, the parameters and the base URL are sensitive

pkg/blockatlas/client.go
Pantani: in most of the cases in the top-level, inside the handler API response

pkg/storage/redis/hmap.go
Pantani: the package storage is not in charge to log something, this is only a wrapper to Redis database, he doesn't know about the log package. The callers of this package need to handle the error and log them

arein: Ok just want to make sure that the consumer of this is logging `value` so we can determine what is passing invalid JSON.

pkg/blockatlas/client.go
arein: Ok thx for explaining.

minor and not sure if adhering to code standards: but to avoid regressions here pls consider adding a brief comment mentioning why we're only passing the `path` and not the full `uri` to the consumer

PR #845, opened 2020-02-13
Use default client for post webhook

Pantani:
Use default client for post webhook to collect metrics for sent observers

### DIFF
--- a/observer/dispatcher.go
+++ b/observer/dispatcher.go
@@ -1,16 +1,12 @@
 package observer
 
 import (
-	"bytes"
-	"encoding/json"
 	"github.com/trustwallet/blockatlas/pkg/blockatlas"
 	"github.com/trustwallet/blockatlas/pkg/errors"
 	"github.com/trustwallet/blockatlas/pkg/logger"
-	"net/http"
 )
 
 type Dispatcher struct {
-	Client http.Client
 }
 
 type DispatchEvent struct {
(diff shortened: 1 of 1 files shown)


catenocrypt:
👍

observer/dispatcher.go
```diff
@@ -29,23 +25,20 @@ func (d *Dispatcher) dispatch(event Event) {
...
 		"txID":    event.Tx.ID,
 	}
-	go d.postWebhook(webhook, txJson, logParams)
+	go d.postWebhook(webhook, action, logParams)
 	logger.Info("Dispatching webhooks...", logger.Params{"webhook": webhook}, logParams)
 }
 
-func (d *Dispatcher) postWebhook(hook string, data []byte, logParams logger.Params) {
-	_, err := d.Client.Post(hook, "application/json", bytes.NewReader(data))
+func (d *Dispatcher) postWebhook(hook string, data interface{}, logParams logger.Params) {
```
vikmeup:
```suggestion
func (d *Dispatcher) postWebhook(url string, data interface{}, logParams logger.Params) {
```

observer/dispatcher.go
```diff
@@ -29,23 +25,20 @@ func (d *Dispatcher) dispatch(event Event) {
...
+	client := blockatlas.InitClient(hook)
+	client.Headers["Content-Type"] = "application/json"
+	err := client.Post(nil, "", data)
```
vikmeup: What is `nil` here?

Pantani: because the function doesn't need the return object

EnoRage:
Is deprecated

Issue #715, opened 2020-01-16
Split implementation of the protocol into separate files


vikmeup:
## Summary of Bug

Currently all implementation for blockchain platform contain all protocol implementation, very hard to navigate and support

Tron has a good example of the structure (https://github.com/trustwallet/blockatlas/tree/master/platform/tron)

```
api.go
api_test.go
client.go
model.go
stake.go
stake_test.go
```

But this could be improve to have `block.go`, `transactions.go` and etc

- [x] Update https://developer.trustwallet.com/blockatlas/newblockchain

vikmeup:
Any reasons to close this issue?

Pantani:
sorry, I forget to link the PR

https://github.com/trustwallet/blockatlas/pull/802

vikmeup:
It was not just stake related problem, it's overall problem to define proper structure for all the files, transactions, block parsing, and etc

Pantani:
> But this could be improve to have block.go, transactions.go and etc

The most of coins use the same models and the normalize for block and transaction API, I don't think is a good idea to split these APIs, which can be more confused. My suggestion is split only no related things, like the ZNS lookup

vikmeup:
Can you define proper files structure for platforms and other services and let's have one guide for having correct file structure

Pantani:
yes, I going to update all docs today

PR #847, opened 2020-02-13
Split platform files per API implementation

Pantani:
Split platform files per API implementation:

- `base.go` for platform base implementation
- `transaction.go` for tx and token tx API's
- `block.go` for block observer
- `stake.go` for naming service API's
- `token.go` for Token API's
- `collection.go` for Collection API's
- `domain.go` for naming service API's

closes https://github.com/trustwallet/blockatlas/issues/715


vikmeup:
address.go? What does this mean? transaction.go or transactions.go would be self explanatory 

What is xpub addresses?

vikmeup:
Why base? Is is transaction or block API?

Pantani:
> address.go? What does this mean? transaction.go or transactions.go would be self explanatory
> 
> What is xpub addresses?

https://github.com/trustwallet/blockatlas/blob/master/pkg/blockatlas/api.go#L45

Pantani:
> Why base? Is is transaction or block API?

Is the both, transactions and block, this is the base methods all coins need to implement before implementing other features

vikmeup:
Block is not requires, many blockchains don't have it implemented

vikmeup:
> But I think is weird as well, do you have another suggestion for a name?

transaction.go is a good name. all it does - return transaction list

vikmeup:
@Pantani can you rebase? let's get this merged

Pantani:
> Looks good! much more organized
> 
> You used to have `custom.go` - was not able to find that in the changes

The only custom route we have now is transactions for bitcoin, I put the `v1/xpub` and `v1/address` inside `transaction.go` to be more clear, but I going to put inside the developer documentation in case we need custom routes, create a `custom.go` file

Issue #834, opened 2020-02-12
Improve handling for inactivate, outdated validators


vikmeup:
## Summary of Bug

Currently we get `v2/tezos/staking/delegations/tz1WDujRWCYjLBDfZieXW6insg5EUbg1rCRK`
```
{
   "code":500,
   "error":{
      "error":"Validator not found: Unable to fetch delegations list",
      "meta":{
         "Balance":"5383274",
         "Delegate":"tz1egbN6RK2bM5vt4aAZw6r9j4nL8z49bPdS"
      }
   }
}
```

This is caused by moving to a new address: https://github.com/trustwallet/assets/commit/f2948f6043cbfb6a5e3a2f81449975adc6d2e2ec#diff-7a2d5e4961dc159a8eff1a77f3f9df70, 
New address: https://github.com/trustwallet/assets/blob/f2948f6043cbfb6a5e3a2f81449975adc6d2e2ec/blockchains/tezos/validators/list.json#L186

## Expected Behavior

- Return list of current delegations
- Do not include outdated and inactive validators into `v2/tezos/staking/validators`

This also affect Cosmos validators, if they become inactive.

Pantani:
@vikmeup If the validator isn't inside the validators map can wee show only the address, like this:

```
{
  "delegations": [
    {
      "delegator": {
        "id": "tz1egbN6RK2bM5vt4aAZw6r9j4nL8z49bPdS",
        "status": false
      },
      "value": "5383274",
      "status": "active"
    }
  ],
  "balance": "5383274",
  "address": "tz1WDujRWCYjLBDfZieXW6insg5EUbg1rCRK",
  "coin": {
    "coin": 1729,
    "symbol": "XTZ",
    "name": "Tezos",
    "decimals": 6
  },
  "details": {
    "reward": {
      "annual": 6.09
    },
    "locktime": 0,
    "minimum_amount": "0",
    "type": "delegate"
  }
}
```

vikmeup:
How does this solve an issue? Why it’s not available in the validators map? 

Reason those questions to find a solution to the problem.

Pantani:
I going to add the validator, but to avoid we show an error when the validator not found, we can return the delegation without the validator info, it's not to solve this issue, to improve the feature

Pantani:
My plan is to use a flag to shows if the validator is outdated or inactive, and we can set this flag without remove them

vikmeup:
[...]

PR #850, opened 2020-02-13
improve response for unactive stake delegation

Pantani:
closes https://github.com/trustwallet/blockatlas/issues/834

- Split services into different files: `info.go` for coin info and `stake.go` for stake methods
- Move stake API handlers to `stake.go` file and remove `delegations.go` file
- Add logic for outdated validators
- Avoid algorithms O(n^2).
- Move asset URL to config file


vikmeup:
Can you split logical peaces into separate commits? I was not able to find easily `Add logic for outdated validators` and how does it solve that problem

Pantani:
> Can you split logical peaces into separate commits? I was not able to find easily `Add logic for outdated validators` and how does it solve that problem

We already have a flag for disabled validators, but we don't use it in the right way. This is flag is the same to remove a validator from the list, after this PR the flag only removes the validator from the validator list endpoint.

I going to split in small PR's

PR #860, opened 2020-02-14
Remove unactive delegate only for stake list

Pantani:
# Headline

Remove inactive delegate only for stake list and keep to fetch  for delegations

## Issue

Validators can change the address or be outdated, but the user can already be staked in this validator

## Solution

Use the `disabled` flag to remove the validator from validator list, but we can show if we need to show delegation info for the user

closes https://github.com/trustwallet/blockatlas/issues/834

### DIFF
--- a/services/assets/stake.go
+++ b/services/assets/stake.go
@@ -23,40 +23,53 @@ func requestValidatorsInfo(coin coin.Coin) (AssetValidators, error) {
 }
 
 func GetValidatorsMap(api blockatlas.StakeAPI) (blockatlas.ValidatorMap, error) {
-	validators, err := GetValidators(api)
+	assets, validators, err := GetValidators(api)
 	if err != nil {
 		return nil, err
 	}
-	return validators.ToMap(), nil
+	results := normalizeValidators(assets, validators, api.Coin(), false)
+	return results.ToMap(), nil
 }
 
-func GetValidators(api blockatlas.StakeAPI) (blockatlas.StakeValidators, error) {
+func GetActiveValidators(api blockatlas.StakeAPI) (blockatlas.StakeValidators, error) {
+	assets, validators, err := GetValidators(api)
+	if err != nil {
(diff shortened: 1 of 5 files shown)


services/assets/stake.go
```diff
@@ -23,40 +23,53 @@ func requestValidatorsInfo(coin coin.Coin) (AssetValidators, error) {
...
-	results := normalizeValidators(validators, assetsValidators, api.Coin())
-	sort.Slice(results, func(i, j int) bool {
-		return results[i].Details.Reward.Annual > results[j].Details.Reward.Annual
-	})
-	return results, nil
+	return assetsValidators, validators, nil
 }
 
-func normalizeValidators(validators []blockatlas.Validator, assets AssetValidators, coin coin.Coin) blockatlas.StakeValidators {
+func normalizeValidators(assets AssetValidators, validators []blockatlas.Validator, coin coin.Coin, onlyActive bool) blockatlas.StakeValidators {
```
vikmeup: can we remove `onlyActive` from this function? if that's possible. function itself should not control this option

for where it's needed just do filter by `onlyActive`.

vikmeup:
Thanks for clean PR, much easier to follow adjusted logic

services/assets/stake.go
Pantani: we need to create another normalize function only for active validator and other for all because  this method this is the only one knows about the asset and the parameter disabled, this parameter doesn't to the top level

vikmeup: Maybe we should expose that status somehow? what other solutions other there?

Pantani: Create a new parameter inside the StakeValidator to pass this to the top lvl and filter that

Pantani: or we can remove the disabled validators before normalize

Pantani: now looks good!

PR #862, opened 2020-02-14
Fix makefile for AZP

Pantani:
The AZP machine doesn't support override command :(


vikmeup:
Should be fixed on master

Issue #116, opened 2019-05-19
Return transactions in descending order instead query max limit


kolya182:
XRP provides option to return transactions in descending order, but apparently API does not work, i have created an issue for it here https://github.com/ripple/rippled-historical-database/issues/198
Currently we query 200 transactions and sort them in descending order https://github.com/TrustWallet/blockatlas/blob/927df068255da649d1b0a71f0eb3c11b3cccc18c/platform/ripple/client.go#L21

If address has over 200 transactions (max possible returned by XRP api is 1000) we will not be able to return the latest.

https://data.ripple.com/v2/accounts/rGig91TmYtssYz8umHGtF6A22gBtwgLxHn/transactions?type=Payment&result=tesSUCCESS&limit=1&descending=true

Waiting for issue to be resolved on Ripple side.

https://xrpl.org/data-api.html#api-method-reference

PR #867, opened 2020-02-14
[RIPPLE] Order tx by desc data and parse failed transactions

Pantani:
- Order transaction by desc data 
- Parse failed transactions

closes https://github.com/trustwallet/blockatlas/issues/116

### DIFF
--- a/platform/ripple/api.go
+++ b/platform/ripple/api.go
@@ -84,7 +84,7 @@ func NormalizeTx(srcTx *Tx) (blockatlas.Tx, bool) {
 	}
 
 	status := blockatlas.StatusCompleted
-	if srcTx.Meta.TransactionResult != tesSUCCESS {
+	if srcTx.Meta.TransactionResult != transactionResultSuccess {
 		status = blockatlas.StatusFailed
 	}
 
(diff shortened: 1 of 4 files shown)


platform/ripple/model.go
```diff
@@ -4,6 +4,15 @@ import (
 	"github.com/trustwallet/blockatlas/pkg/blockatlas"
 )
 
+type TransactionType string
+type TransactionResult string
+
+const (
+	tesSUCCESS TransactionResult = "tesSUCCESS"
```
vikmeup:
```suggestion
	transactionResultSuccess TransactionResult = "tesSUCCESS"
``` ?

File: platform/ripple/api.go [hunk 2]
```diff
@@ -82,18 +79,24 @@ func NormalizeTx(srcTx *Tx) (blockatlas.Tx, bool) {
 		return blockatlas.Tx{}, false
 	}
 
-	if srcTx.Payment.TransactionType != "Payment" {
+	if srcTx.Payment.TransactionType != transactionPayment {
 		return blockatlas.Tx{}, false
 	}
 
+	status := blockatlas.StatusCompleted
+	if srcTx.Meta.TransactionResult != tesSUCCESS {
```
vikmeup: what other possible cases there is? to avoid defaulting to failed

Pantani: all other codes are errors, only one can be a success is the ter Codes (Retry), but another transaction going to be applied in another block

https://xrpl.org/transaction-results.html

PR #881, opened 2020-02-17
Set Max and Min block time for observer

Pantani:
- Remove the logic to divide the block time by 4
- Add max block time, 30s
- Reduce the minimum block time to 3s

### DIFF
--- a/cmd/platform_observer/main.go
+++ b/cmd/platform_observer/main.go
@@ -32,8 +32,12 @@ func main() {
 		logger.Fatal("No APIs to observe")
 	}
 
-	minInterval := viper.GetDuration("observer.min_poll")
 	backlogTime := viper.GetDuration("observer.backlog")
+	minInterval := viper.GetDuration("observer.min_poll")
+	maxInterval := viper.GetDuration("observer.max_poll")
+	if minInterval >= maxInterval {
+		logger.Fatal("minimum block polling interval cannot be greater or equal than maximum")
+	}
 
 	var wg sync.WaitGroup
 	wg.Add(len(platform.BlockAPIs))
@@ -41,10 +45,10 @@ func main() {
 	for _, api := range platform.BlockAPIs {
 		coin := api.Coin()
 		blockTime := time.Duration(coin.BlockTime) * time.Millisecond
-		pollInterval := blockTime / 4
-
-		if pollInterval < minInterval {
-			pollInterval = minInterval
+		if blockTime < minInterval {
+			blockTime = minInterval
+		} else if blockTime > maxInterval {
+			blockTime = maxInterval
 		}
 
 		// Stream incoming blocks
@@ -59,7 +63,7 @@ func main() {
 		stream := observer.Stream{
 			BlockAPI:     api,
 			Tracker:      cache,
-			PollInterval: pollInterval,
+			PollInterval: blockTime,
 			BacklogCount: backlogCount,
 		}
 		blocks := stream.Execute(context.Background())
@@ -80,7 +84,7 @@ func main() {
 
 		logger.Info("Observing", logger.Params{
 			"coin":     coin,
-			"interval": pollInterval,
+			"interval": blockTime,
 			"backlog":  backlogCount,
 		})
 	}
--- a/cmd/platform_observer/main.go
+++ b/cmd/platform_observer/main.go
@@ -10,6 +10,7 @@ import (
 	"github.com/trustwallet/blockatlas/platform"
 	"github.com/trustwallet/blockatlas/storage"
 	"sync"
+	"time"
 )
 
 const (
@@ -44,7 +45,10 @@ func main() {
 
 	for _, api := range platform.BlockAPIs {
 		coin := api.Coin()
-		pollInterval := numbers.GetInterval(coin.BlockTime, minInterval, maxInterval)
+		interval := time.Duration(coin.BlockTime) * time.Millisecond
+		pMin := numbers.Max(minInterval.Nanoseconds(), interval.Nanoseconds())
+		pMax := numbers.Min(int(maxInterval.Nanoseconds()), int(pMin))
+		pollInterval := time.Duration(pMax)
 
 		// Stream incoming blocks
 		var backlogCount int
--- a/cmd/platform_observer/main.go
+++ b/cmd/platform_observer/main.go
@@ -33,8 +33,8 @@ func main() {
 	}
 
 	backlogTime := viper.GetDuration("observer.backlog")
-	minInterval := viper.GetDuration("observer.min_poll")
-	maxInterval := viper.GetDuration("observer.max_poll")
+	minInterval := viper.GetDuration("observer.block_poll.min")
+	maxInterval := viper.GetDuration("observer.block_poll.max")
 	if minInterval >= maxInterval {
 		logger.Fatal("minimum block polling interval cannot be greater or equal than maximum")
 	}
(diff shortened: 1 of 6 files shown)


cmd/platform_observer/main.go
```diff
@@ -32,19 +32,23 @@ func main() {
...
 	wg.Add(len(platform.BlockAPIs))
 
 	for _, api := range platform.BlockAPIs {
 		coin := api.Coin()
 		blockTime := time.Duration(coin.BlockTime) * time.Millisecond
-		pollInterval := blockTime / 4
-
-		if pollInterval < minInterval {
-			pollInterval = minInterval
+		if blockTime < minInterval {
```
vikmeup: did you consider moving this into independent function? 

`func (p *Platform)GetBlockPoolInternal() { }` 

easier to test and read.

Please try to not just fix bugs - always improve code

cmd/platform_observer/main.go
```diff
@@ -80,7 +84,7 @@ func main() {
 
 		logger.Info("Observing", logger.Params{
 			"coin":     coin,
-			"interval": pollInterval,
+			"interval": blockTime,
```
vikmeup: `blockTime` wrong name - block time is not equal to pooling internal

config.yml
```diff
@@ -22,7 +22,9 @@ observer:
   # Redis Subscription Database
   redis: redis://redis:6379
   # Smallest possible block polling interval
-  min_poll: 250ms
```
vikmeup: did you consider? 

```
block_poll
    min: 3s
    max: 30s
```

cmd/platform_observer/main.go
```diff
@@ -59,7 +63,7 @@ func main() {
 		stream := observer.Stream{
 			BlockAPI:     api,
 			Tracker:      cache,
-			PollInterval: pollInterval,
+			PollInterval: blockTime,
```
vikmeup: same as https://github.com/trustwallet/blockatlas/pull/881/files#r380305169

cmd/platform_observer/main.go
```diff
@@ -32,19 +32,23 @@ func main() {
 		logger.Fatal("No APIs to observe")
 	}
 
-	minInterval := viper.GetDuration("observer.min_poll")
 	backlogTime := viper.GetDuration("observer.backlog")
+	minInterval := viper.GetDuration("observer.min_poll")
+	maxInterval := viper.GetDuration("observer.max_poll")
+	if minInterval >= maxInterval {
+		logger.Fatal("minimum block polling interval cannot be greater or equal than maximum")
```
vikmeup: why log this?

cmd/platform_observer/main.go
vikmeup: Add tests for this

cmd/platform_observer/main.go
Pantani: If the min is greater the min, we going to trigger a fatal error to stop the application, because can bug the logic

config.yml
```diff
@@ -19,16 +19,16 @@ observer:
   enabled: false
   # TODO : Add description
   auth: test
-  # Redis Subscription Database
-  redis: redis://redis:6379
```
Pantani: it is duplicate, now we have `storage.redis`

cmd/platform_observer/main.go
Pantani: `func (p *Platform)GetBlockPoolInternal() { }` per platform is not a good idea, we need to implement that inside all platforms to do the same thing. we can create a function for this, but we need to always get an env var per platform, now we get only one time for all. Another way is to create a function like `func poolInterval(blockTime, min, max time.Duration) time.Duration`

cmd/platform_observer/main.go
vikmeup: Maybe you can write code so you can’t break logic? 

There is a reason why min, max function exist, to avoid such code

config.yml
vikmeup: Why touching this code in this PR?

vikmeup: This requires changes in deployment pipeline.

cmd/platform_observer/main.go
vikmeup: 👍 do they right thing

config.yml
Pantani: I going remove and change in another PR, we use `storage.redis` now

pkg/numbers/number.go
```diff
@@ -69,3 +70,13 @@ func SliceAtoi(sa []string) ([]int, error) {
 	}
 	return si, nil
 }
+
+func GetInterval(blockTime int, minInterval, maxInterval time.Duration) (pollInterval time.Duration) {
```
vikmeup: if function name is `GetInterval`, why does it do anything related to `poll`?

pkg/numbers/number.go
```diff
@@ -71,12 +71,12 @@ func SliceAtoi(sa []string) ([]int, error) {
...
-func GetInterval(blockTime int, minInterval, maxInterval time.Duration) (pollInterval time.Duration) {
-	pollInterval = time.Duration(blockTime) * time.Millisecond
-	if pollInterval < minInterval {
-		pollInterval = minInterval
-	} else if pollInterval > maxInterval {
-		pollInterval = maxInterval
+func GetInterval(value int, minInterval, maxInterval time.Duration) (interval time.Duration) {
+	interval = time.Duration(value) * time.Millisecond
+	if interval < minInterval {
+		interval = minInterval
```
vikmeup: did you consider using `max`, `min` functions for this? would be much more readable

pkg/numbers/number.go
```diff
@@ -71,12 +71,12 @@ func SliceAtoi(sa []string) ([]int, error) {
...
+	} else if interval > maxInterval {
+		interval = maxInterval
```
vikmeup: not sure why this needs to be part of numbers package

pkg/numbers/number.go
Pantani: yes! but we are comparing time. interval, if we use the Max or Min function we need to do a lot of casting

pkg/numbers/number.go
Pantani: I think we need to keep together with the max and min function, we can move all of this to other packages or keep inside the numbers package. what's your suggestion?

pkg/numbers/number.go
vikmeup: There is two pieces:
1) calculate internal (observer specific)
2) get min internal (just min function? why complicated with such design?)

I do not understand why it needs a separate function in the numbers package?

Pantani: please, give me your suggestion, because it's not clear whats you are trying to say and we can spend less time fixing a simple thing. It's good for me this way, works, is testable, clean and  I don't wanna put this logic inside the observer's main package (cmd). 

The max and min function is used for `int`, we are comparing `time.Duration`

vikmeup: https://en.wikipedia.org/wiki/Single_responsibility_principle

cmd/platform_observer/main.go
```diff
@@ -44,7 +45,10 @@ func main() {
 
 	for _, api := range platform.BlockAPIs {
 		coin := api.Coin()
-		pollInterval := numbers.GetInterval(coin.BlockTime, minInterval, maxInterval)
+		interval := time.Duration(coin.BlockTime) * time.Millisecond
+		pMin := numbers.Max(minInterval.Nanoseconds(), interval.Nanoseconds())
```
vikmeup: can you add test to this?

observer/observer.go
```diff
@@ -106,3 +108,10 @@ func inferUtxoValue(tx *blockatlas.Tx, address string, coinIndex uint) {
 		}
 	}
 }
+
+func GetInterval(value int, minInterval, maxInterval time.Duration) time.Duration {
```
vikmeup: `value int` - what does it mean in this context?
`GetInterval` - is this generic internal of block internal ? 

I'm not able to easily follow what does the function does.

observer/observer_test.go
```diff
@@ -247,3 +248,50 @@ func Test_inferUtxoValue(t *testing.T) {
...
+	}
+	tests := []struct {
+		name string
+		args args
+		want time.Duration
+	}{
+		{
+			"test minimum",
+			args{
+				blockTime:   100,
```
vikmeup: why it's called correctly here as blocktime and not as here: https://github.com/trustwallet/blockatlas/pull/881/commits/496e331b5d395e6eeb8bfc5d7ddf35905c04a36c#diff-483a22458f1a7ae9c358127c7c4f8d7bR112 ?

PR #890, opened 2020-02-19
Create a package to client methods

Pantani:
# Headline

Move client methods to own package to be clear


EnoRage:
What are the reasons to do it?

Pantani:
> What are the reasons to do it?

@EnoRage 
All reusable method is inside the Blockatlas package, let's keep only the Blockatlas stuff inside them, and we can create helpers package for other stuff, what do you think?

EnoRage:
cannot see any sense for it

PR #910, opened 2020-02-22
Fix collections

EnoRage:
Fix an issue with an empty http client during v3/collections fetch.
The main problem was because the Init function must initialize collections api for all ETH coins. 
It is important to rewrite the logic of collections api, so there will be only suitable for collections tokens (ETH only for now)

### DIFF
--- a/platform/platform.go
+++ b/platform/platform.go
@@ -31,6 +31,8 @@ import (
 	"github.com/trustwallet/blockatlas/platform/zilliqa"
 )
 
+const emptyParam = ""
+
 func GetVar(name string) string {
 	return viper.GetString(name)
 }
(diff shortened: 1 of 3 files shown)


platform/platform.go
```diff
@@ -31,6 +31,8 @@ import (
 	"github.com/trustwallet/blockatlas/platform/zilliqa"
 )
 
+const emptyParam = ""
```
vikmeup: Any reason it’s called this way? If it’s an emptyString

platform/platform.go
```diff
@@ -85,13 +87,13 @@ func getPlatformMap() blockatlas.Platforms {
...
 		coin.Qtum().Handle:         bitcoin.Init(coin.QTUM, GetApiVar(coin.QTUM)),
-		coin.Gochain().Handle:      ethereum.Init(coin.GO, GetApiVar(coin.GO), GetRpcVar(coin.GO)),
-		coin.Thundertoken().Handle: ethereum.Init(coin.TT, GetApiVar(coin.TT), GetRpcVar(coin.TT)),
-		coin.Classic().Handle:      ethereum.Init(coin.ETC, GetApiVar(coin.ETC), GetRpcVar(coin.ETC)),
-		coin.Poa().Handle:          ethereum.Init(coin.POA, GetApiVar(coin.POA), GetRpcVar(coin.POA)),
-		coin.Callisto().Handle:     ethereum.Init(coin.CLO, GetApiVar(coin.CLO), GetRpcVar(coin.CLO)),
-		coin.Wanchain().Handle:     ethereum.Init(coin.WAN, GetApiVar(coin.WAN), GetRpcVar(coin.WAN)),
-		coin.Tomochain().Handle:    ethereum.Init(coin.TOMO, GetApiVar(coin.TOMO), GetRpcVar(coin.TOMO)),
-		coin.Ethereum().Handle:     ethereum.InitWitCollection(coin.ETH, GetApiVar(coin.ETH), GetRpcVar(coin.ETH), GetVar("ethereum.collections_api"), GetVar("ethereum.collections_api_key")),
+		coin.Gochain().Handle:      ethereum.Init(coin.GO, GetApiVar(coin.GO), GetRpcVar(coin.GO), emptyParam, emptyParam),
```
vikmeup: Any chance to create separate initializer without empty parameters?

EnoRage: We need it init Collection API only with ETH otherwise

platform/platform.go
EnoRage: Param to init Collection API, gives a bit more details about behavior of function from my point of view

platform/platform.go
EnoRage: https://github.com/trustwallet/blockatlas/issues/912 - will be solved later

vikmeup: Any chance just create separate initializer for them?

PR #920, opened 2020-02-24
[Observer] Add new provider to send dispatched events, add RabbitMQ

EnoRage:


observer/dispatcher.go
```diff
@@ -11,11 +12,15 @@ import (
 
 type Dispatcher struct {
 	Client http.Client
+
+	// amqp or http
+	DispatchProtocol string
```
prazd: I think it's better to use type
```
type DispatchProtocol string
const HTTP DispatchProtocol = "http"
const AMQP DispatchProtocol = "amqp"
```

observer/dispatcher.go
```diff
@@ -25,25 +30,37 @@ func (d *Dispatcher) Run(events <-chan Event) {
...
 		"coin":    event.Subscription.Coin,
 		"txID":    event.Tx.ID,
 	}
-	go d.postWebhook(webhook, txJson, logParams)
-	logger.Info("Dispatching webhooks...", logger.Params{"webhook": webhook}, logParams)
+	switch d.DispatchProtocol {
+	case "http":
+		go d.postWebhook(webhook, txJson, logParams)
+	case "amqp":
+		go d.postMessageToQueue(webhook, txJson, logParams)
```
prazd: And use DispatchProtocol type here

Issue #927, opened 2020-02-25
[Platform] Migrate all mocking test JSON code into separate files


vikmeup:
## Summary of Bug

Currently we store all JSON code inside go lang files (https://github.com/trustwallet/blockatlas/blob/master/platform/ethereum/collection_test.go#L15), which become messy and hard to write new test, better to store JSON into separate files

EnoRage:
Can you please share your thoughts about it @kolya182 ?
I suggest just to use the consistent pattern for all platform tests and not using separete files. I did not saw many cases like this in other Golang repos like https://github.com/kubernetes/kubernetes

EnoRage:
@catenocrypt are you ok with that? The main idea that was proposed by @vikmeup is to use .json files for mocked responses to prevent duplicates. Another reason use to simplify the way how we can reuse that files for other tests

catenocrypt:
Some notes:
- This applies to .go test files, but also the newly-introduced .js test files (mocked with dyson)
- Response could be stored in json files, in a folder structure, where subfolders identify the API, the path, and query parameters.  E.g.
`eth/api/v1%C6addres%C60xf01010101%C7details%C8tx.json' (made-up example)
where eth/api identifies the api endpoint (from config file), filename is URL-encoded full path
- It is even possible to automate updating/creating these json files, by retrieving the responses from real endpoints.

EnoRage:
Do we have any progress on that? @catenocrypt

PR #929, opened 2020-02-26
Add transaction direction

kolya182:
Add transaction direction, which was outgoing by default + tests

### DIFF
--- a/platform/vechain/transaction_test.go
+++ b/platform/vechain/transaction_test.go
@@ -27,15 +27,16 @@ const trxId = `{
 }`
 
 var expectedTransfer = blockatlas.Tx{
-	ID:     "0x702edd54bd4e13e0012798cc8b2dfa52f7150173945103d203fae26b8e3d2ed7",
-	Coin:   coin.VET,
-	From:   "0xB5e883349e68aB59307d1604555AC890fAC47128",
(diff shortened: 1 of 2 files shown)


platform/vechain/transaction_test.go
```diff
@@ -27,15 +27,16 @@ const trxId = `{
 }`
 
 var expectedTransfer = blockatlas.Tx{
-	ID:     "0x702edd54bd4e13e0012798cc8b2dfa52f7150173945103d203fae26b8e3d2ed7",
```
EnoRage: Can I kindly ask to create a separate JSON file for tests? According to https://github.com/trustwallet/blockatlas/issues/927 
We need to use separate files for mocks

EnoRage: Example 
https://github.com/trustwallet/blockatlas/commit/f3fd50b098e650a1938aadbbda030611bd586cdb

kolya182:
> Let's check that `len(event.Topics[1]) > 3` before `event.Topics[1]` or `event.Topics[2]` to prevent any possibilty to have an error

We do have it already https://github.com/trustwallet/blockatlas/blob/master/platform/vechain/transaction.go#L89

PR #930, opened 2020-02-26
[Do not read yet] Change http error response and RenderError function

EnoRage:


EnoRage:
Not yet finished

EnoRage:
Too much outdated, will be resolved again in the new PR

PR #953, opened 2020-03-03
Internal init changes, add optinal MQ prefetch count 

EnoRage:


mq/mq.go
```diff
@@ -67,11 +70,17 @@ func (q Queue) RunConsumer(consumer Consumer, cache storage.Addresses) {
 		return
 	}
 
+	if PrefetchCount < minPrefetchCount {
+		logger.Info("Change prefetch count to default")
+		PrefetchCount = defaultPrefetchCount
+	}
```
arein: How do you feel about `panic`ing here?

If not - consider to make it a `logger.warn` and to add the `PrefetchCount` and `minPrefetchCount` to the log statement so it's easier to fix

EnoRage: I though about panic, but if we will have a broken container on production with auto restart - it is worse than having small optimized prefetch count

EnoRage: Will change to warn

EnoRage:
I will bring Docker tests for Observer, dangerous to make PRs without them

PR #886, opened 2020-02-18
[Test] Sample for mock for external services

catenocrypt:
For functional test that are isolated from the eventual errors within the external services, mocks 
can be used to simulate the external services.

A simple mock framework **Dyson** is used for mocking (https://github.com/webpro/dyson).
Simple Javascript files can be used to specify the responses, it is easy to create dynamic responses, easy to install and start dyson, so it can be used in CI as well.  One dyson instance can handle many API routes.

I have created two sample API mocks; one for FIO name lookup API (POST), and one for binance explorer transaction query (GET).  They are set up to return data used by platform-api functional tests.  Config filed for dyson are located here:   `mock/ext-api-dyson`.

Testing:
```
Install Dyson:
make install-dyson       or         npm install -g dyson
Start dyson with the mocks for external APIs:
make start-mock-dyson            or         dyson  mock/ext-api-dyson
Start platform-api with config for mocked services:
make start-platform-api-mocked            or       
go build -o platform-api-bin cmd/platform_api/main.go  && ./platform-api-bin -p 8420 -c configmock.yml
Start functional tests
make go-functional
Other test can be also executed (e.g. postman API tests)
Or the whole thing in one command:
make go-functional-mocked
```

PR #963, opened 2020-03-13
Base infra for mocked external services, mocked functional tests.

catenocrypt:
For functional test isolated from errors of the external services, mocks can be used to simulate external services.

A simple mock framework **Dyson** is used for mocking (https://github.com/webpro/dyson).
Simple Javascript files can be used to specify the responses, it is easy to create dynamic responses, easy to install and start dyson, so it can be used in CI as well.  One dyson instance can handle many API routes.

I have created two sample API mocks; one for FIO name lookup API (POST), and one for binance explorer transaction query (GET).  They are set up to return data used by platform-api functional tests.  Config files for dyson are located here:   `mock/ext-api-dyson`.

The external service expoints are exchanged now simply by a duplicated configmock.yml file -- this should be improved, as it creates some undesirable duplication.

Testing:
```
Install Dyson:
make install-dyson       or         npm install -g dyson

Start dyson with the mocks for external APIs:
make start-mock-dyson            or         dyson  mock/ext-api-dyson

Start platform-api with config for mocked services:
make start-platform-api-mocked            or       
go build -o platform-api-bin cmd/platform_api/main.go  && ./platform-api-bin -p 8420 -c configmock.yml

Start functional tests
make go-functional

Other test can be also executed (e.g. postman API tests)

Or the whole thing in one command:
**make go-functional-mocked**
```


EnoRage:
It is a bit hard to pick the suitable mock provider

We already use  github.com/vektra/mockery at Watchmarket.

The main + of it is that we can use Golang structures there. Also it is better when all the environment is with one lang 

We need to make a research 

I want to remind that there is a Watchmarket project and it is important to have the common tests framework and environment if it is possible

Currently we use:
docker tests for integration tests
mockery for mocking data through unit tests
newman for testing with real endpoints 

There is no strict restriction to use specifically these tools, but we need to make a decision for the whole our backend ecosystem

catenocrypt:
I am not particularly keen on dyson, this was the best choice I found a few weeks ago.  Definitely the more consistent the tooling is, the better.  I will check out mockery as well.

My criteria were/are:
- easy to download/install  (if mockery is go-based, should be no problem)
- should be easy to specify concrete return values and simple dynamic content, of the form 'if input is A, then response is B, ...'.  Some tools based on api specifications or response files are not suitable for this.
- should mock on HTTP API level, run in a separate process.

I have compared some alternatives and sought feedback (in #886 ).  Let's try afresh.

catenocrypt:
Updated -- criteria for mock tool for external services:
- easy to download/install
- should mock on HTTP API level, run in a separate process (to avoid mixing of test and product code).
- should be easy to specify concrete return values and simple dynamic content, of the form 'if input is A, then response is B, ...'. Some tools based on api specifications or response files are not suitable for this.
- should run in a local process, not hosted (paid nor free).  Hosted solutions are convenient, but there's less control (e.g. we need different version for each PR).

catenocrypt:
The following invocation could be added to the CI build right away (once the PR is accepted):
```
make newman-mocked test=domain host=http://localhost:8420
```

EnoRage:
It is great solution, hope it will help us to cover all the external APIs soon. That is one of the main issues now

Issue #966, opened 2020-03-16
[Observer] Refactor notifications observer


EnoRage:
Current observer architecture have several issues:
1. Channel hell
2. Not isolated i/o operations => 
3. Hard to test and to isolate something, we need almost full flow (read readme.md) to test it

New solution:

Observer will have 2 modules:

1. [Block parser] Fetching data from external api, parse it to the []blockatlas.Block and push to the MQ queue
2. [Transactions notifier] Get Block from MQ, parse transactions, find users that have that transaction at Redis, send transaction event to the next queue 

It will allow us to mock all the data for 2 much more easily and be sure that if block was parsed and is at queue => 100% it will be checked for users notifications (MQ + ack)

hewigovens:
current approach is polling, but some blockchain / nodes have the ability to subscribe via websocket, maybe we can add a generic subscriber as well?

EnoRage:
Websockets are not stable solution, even Ethereum does not have stability with them, we loose blocks using them on others projects, also polling strategy is more convenient if we are batching blocks and batching transactions also. I do not think that something will significantly changes if we will add them

EnoRage:
https://github.com/trustwallet/blockatlas/pull/968

PR #968, opened 2020-03-17
[Observer] Start rewriting observer service: separate service - add observer_parser and observer_notifier

prazd:


EnoRage:
#966

Issue #959, opened 2020-03-11
[Platform] Remove all real endpoints from integration and functional tests


EnoRage:
Some endpoint like fiotestnet (testnet) can broke pipeline. We need to mock each response and do not use real ones. We have status tests for that. 

- [x] domains
- [x] tokens
- [x] staking
- [x] collection
- [x] transaction (all the coins)

EnoRage:
https://github.com/trustwallet/blockatlas/pull/963
Is a great start

EnoRage:
https://github.com/trustwallet/blockatlas/pull/971
Transactions are done. @catenocrypt am I right?

catenocrypt:
Correct, transactions and domain are done.

EnoRage:
@catenocrypt added staking via #998

EnoRage:
#1066 collections were done by @catenocrypt

EnoRage:
Observer mocks needs to be implemented together with more detailed integration test

PR #971, opened 2020-03-21
Mocked transaction history tests.

catenocrypt:
More tests with mocked external services.
See #959 .


catenocrypt:
Note: most of the commits appearing in the list have been already merged to master.

catenocrypt:
Transaction tests have been 'mockified', for all coins.

The command `make newman-mocked host=localhost:8420` should be added as a new CI build.

Transaction tests:
- non-mocked: runs in 57 seconds, 14/58 cases fail.
- mocked: runs under 4 seconds, 0 failures.
It even runs with no network connectivity :)

PR #975, opened 2020-03-25
Migration from Redis to Postgres DB

EnoRage:


hewigovens:
@vikmeup I remember Danilo had tried PG? no idea why it didn't work for him

db/models/subscriptions.go
```diff
@@ -4,13 +4,14 @@ import "github.com/jinzhu/gorm"
 
 type Subscription struct {
 	gorm.Model
-	GUID string             `gorm:"primary_key:true"`
+	GUID string             `gorm:"type:varchar(256);unique_index"`
```
vikmeup: let's use 64 here

db/models/subscriptions.go
```diff
@@ -4,13 +4,14 @@ import "github.com/jinzhu/gorm"
...
 	Data []SubscriptionData `gorm:"many2many:subscription_data;association_foreignkey:subscriptionID;foreignkey:GUID"`
 }
 
 type SubscriptionData struct {
 	gorm.Model
-	SubscriptionID string `gorm:"primary_key:true"`
-	Coin           uint
-	Address        string
+	Subscription     Subscription `gorm:"ForeignKey:UserId; not null"`
+	SubscriptionGUID string       `sql:"index"`
```
vikmeup: this would be `SubscriptionID` - as you gonna point to ID and not Guid

db/models/subscriptions.go
```diff
@@ -0,0 +1,17 @@
...
+
+type Subscription struct {
+	gorm.Model
+	GUID string `gorm:"type:varchar(256);unique_index"`
+	Data []SubscriptionData
+}
+
+type SubscriptionData struct {
+	Subscription     Subscription `gorm:"ForeignKey:UserId; not null"`
+	SubscriptionGUID string       `sql:"index"`
```
vikmeup: `SubscriptionGUID` is not needed here I think

cmd/observer_notifier/main.go
```diff
@@ -17,8 +16,8 @@ const (
 )
 
 var (
-	confPath string
-	dbConn   *gorm.DB
+	confPath   string
+	dbInstance *db.Instance
```
vikmeup:
```suggestion
	db *db.Instance
```

EnoRage: using the same naming as package?

PR #976, opened 2020-03-25
[TX] Fix GetTransactionDirection

prazd:


EnoRage:
What is the point of that fix? What is that?

prazd:
The problem was that earlier in GetDirection we used a copy of blockatlas.Tx - https://github.com/trustwallet/blockatlas/blob/3ee6810bd173a5e753ed84ad72c3ec9b087bb7bc/observer/observer.go#L86

But now it is a method that uses *blockatlas.Tx - https://github.com/trustwallet/blockatlas/blob/2190c3b07fa7f5eadbddcba309774c8576e6a693/pkg/blockatlas/tx.go#L267

And if we use pointer on blockatlas.Tx we should use pointer on meta type in switch case - 
https://github.com/trustwallet/blockatlas/blob/2190c3b07fa7f5eadbddcba309774c8576e6a693/pkg/blockatlas/tx.go#L275

PR #979, opened 2020-03-26
Fix for 'make stop' (Makefile, observer pid).

catenocrypt:
Make stop did hang, due to a typo (PID_OBSERVER_NOTIFIER instead of OBSERVER_NOTIFIER).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -169,7 +169,7 @@ ifeq (,$(shell which newman))
 endif
 
 ## newman: Run Postman Newman test, the host parameter is required, and you can specify the name of the test do you wanna run (transaction, token, staking, collection, domain, healthcheck, observer). e.g $ make newman test=staking host=http//localhost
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -169,7 +169,7 @@ ifeq (,$(shell which newman))
 endif
 
 ## newman: Run Postman Newman test, the host parameter is required, and you can specify the name of the test do you wanna run (transaction, token, staking, collection, domain, healthcheck, observer). e.g $ make newman test=staking host=http//localhost
-newman: install-newman start-platform-api
```
catenocrypt: Why remove starting platform?

EnoRage: Because we do not need to run anything here. Host -  is any endpoint, not specially the localhost

catenocrypt: +1

Issue #956, opened 2020-03-10
[Platform] Tezos change API provider


EnoRage:
Try:
tz1VFiQwFL36Ru7EtugLjXVh3dqp9B4KRSAV
https://tezblock.io/account/tz1VFiQwFL36Ru7EtugLjXVh3dqp9B4KRSAV

And also:
tz1foWxaV3VQyWqFbWTERS6YDJjPT6C7jPp8
https://tezblock.io/account/tz1foWxaV3VQyWqFbWTERS6YDJjPT6C7jPp8

PR #980, opened 2020-03-27
Replace current Tezos explorer with TxStats

kolya182:
Close https://github.com/trustwallet/blockatlas/issues/956
- [X] Found couple of issues with explore api, talking to tzstas team to resolve it.
- [X] Rewrite test
- [X] Remove redundant test/mock data in favore unit tests


platform/tezos/client.go
```diff
@@ -1,27 +1,41 @@
...
 )
 
 type Client struct {
 	blockatlas.Request
 }
 
-func (c *Client) GetTxsOfAddress(address string, txType TxType) (txs []Transaction, err error) {
-	err = c.Get(&txs, "v1/"+string(txType), url.Values{"n": {"50"}, "p": {"0"}, "account": {address}})
+func (c *Client) GetTxsOfAddress(address string, txType []string) (txs ExplorerAccount, err error) {
+	var path = fmt.Sprintf("/account/%s/op", address)
```
EnoRage: path :=

platform/tezos/client.go
```diff
@@ -1,27 +1,41 @@
...
+	var status Status
+	err := c.Get(&status, "/status", nil)
+	return status.Indexed, err
 }
 
-func (c *Client) GetBlockByNumber(num int64, txType TxType) (txs []Transaction, err error) {
-	err = c.Get(&txs, "v1/"+string(txType), url.Values{"n": {"50"}, "p": {"0"}, "block": {strconv.Itoa(int(num))}})
-	return
+func (c *Client) GetBlockByNumber(num int64, txType []string) ([]Transaction, error) {
+	var blockOps ExplorerAccount
```
EnoRage: var (
blockOps ExplorerAccount
path = fmt.Sprintf("/account/%d/op", num)
types = strings.Join(txType, ",")
)

platform/tezos/model.go
```diff
@@ -1,94 +1,114 @@
...
+		return blockatlas.StatusError
 	}
-	return t.Delegation.Delegate
 }
 
-func (t *Transaction) Fee() string {
-	if len(t.Tx.Fee) > 0 {
-		return t.Tx.Fee
+func (t *Transaction) ErrorMsg() string {
+	switch t.IsSuccess {
```
prazd: Maybe it’s better not to use a switch here?

platform/tezos/block.go
```diff
@@ -3,36 +3,20 @@ package tezos
...
+		return nil, err
 	}
-	wg.Wait()
-	close(out)
-	srcTxs := make([]Transaction, 0)
-	for r := range out {
-		srcTxs = append(srcTxs, r...)
-	}
-	txs := NormalizeTxs(srcTxs)
+	txs := NormalizeTxs(srcTxs, "")
```
prazd: Empty string does not look good here imho. What do you think?

platform/tezos/block.go
```diff
@@ -3,36 +3,20 @@ package tezos
...
-		go func(txType TxType, num int64, wg *sync.WaitGroup) {
-			defer wg.Done()
-			txs, err := p.client.GetBlockByNumber(num, txType)
-			if err != nil {
-				logger.Error("GetAddrTxs", err, logger.Params{"txType": txType, "num": num})
-				return
-			}
-			out <- txs
-		}(t, num, &wg)
+	txTypes := []string{TxTypeTransaction, TxTypeDelegation}
```
prazd: Is it works synchronous now?
If yes - why we not use concurrency here?

platform/tezos/model.go
```diff
@@ -60,14 +60,13 @@ func (t *Transaction) Status() blockatlas.Status {
...
-		if len(t.Errors) > 0 {
-			return fmt.Sprintf("%s %s", t.Errors[0].ID, t.Errors[0].Kind)
-		}
-		return "transaction error"
+	if t.IsSuccess {
+		return ""
+	} else if len(t.Errors) > 0 {
+		return fmt.Sprintf("%s %s", t.Errors[0].ID, t.Errors[0].Kind)
+	} else {
+		return ""
```
prazd: I think it can be more simplified:
```
func (t *Transaction) ErrorMsg() string {
   if !t.IsSuccess && len(t.Errors) > 0 {
      return fmt.Sprintf("%s %s", t.Errors[0].ID, t.Errors[0].Kind)
  }
  return ""
}
```

platform/tezos/model.go
kolya182: https://github.com/trustwallet/blockatlas/pull/980/commits/64671a7ef3e517b2e52cd479ab40b70b1e56a565

platform/tezos/client.go
kolya182: https://github.com/trustwallet/blockatlas/pull/980/commits/98d7336008094e6550943628bd169de3e32869b1

platform/tezos/client.go
kolya182: https://github.com/trustwallet/blockatlas/pull/980/commits/98d7336008094e6550943628bd169de3e32869b1

platform/tezos/block.go
kolya182: Not elegant, i agree. But so far only one way to determine transaction direction in `Normalize` function.

catenocrypt:
I have updated the mocked unit test to the new API.

PR #992, opened 2020-03-31
Migrate from Trust Ray to blockbook

### LINKED ISSUE #954: [Platform] Explore migration trust-ray => blockbook

hewigovens:
Fixes #954 
- [x] Add generic `EthereumClient` interface
- [x] Add blockbook/ens/collectible/trustray sub packages
- [x] Get token lists
- [x] Add Blockbook mock tests

Non-blocker: 

- [ ] No data field in ContractCall


hewigovens:
I think we have to fork blockbook to finish those left TODOs

catenocrypt:
> I think we have to fork blockbook to finish those left TODOs

Forking blockbook is one thing, but running it (together with a node) is more complicated, no?

hewigovens:
not really, that's exactly what we did for bitcoin, blockbook (indexer) + rpc node (bitcoind) in one instance

kolya182:
Added token list https://github.com/trustwallet/blockatlas/pull/992/commits/1bf18ab7d708a052c11c529b68792a77236c1e36

hewigovens:
I want to test it in non-production environment with iOS app before merging, @EnoRage how to achieve this?

EnoRage:
> I want to test it in non-production environment with iOS app before merging, @EnoRage how to achieve this?

API provided

EnoRage:
I had some issue for TX observer with functions, will need more time to review it

platform/ethereum/blockbook/transaction.go
```diff
@@ -0,0 +1,146 @@
...
+		return blockatlas.StatusError, "Error"
+	case 1:
+		return blockatlas.StatusCompleted, ""
+	default:
+		return blockatlas.StatusError, "Unable to define transaction status"
+	}
+}
+
+func getFrom(srcTx *Transaction) string {
+	if len(srcTx.Vin) > 0 {
```
vikmeup, line 75: should we check for `len(srcTx.Vin[0].Addresses)` as well ?

platform/ethereum/blockbook/transaction.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+func normalizeTxWithAddress(srcTx *Transaction, address, token string, coinIndex uint) blockatlas.Tx {
+	normalized := normalizeTx(srcTx, coinIndex)
+	normalized.Direction = getDirection(address, normalized.From, normalized.To)
+	fillMeta(&normalized, srcTx, address, token, coinIndex)
+	return normalized
+}
+
+func getStatus(specific *EthereumSpecific) (blockatlas.Status, string) {
```
vikmeup, line 61: can you add extension to `EthereumSpecific` instead as `EthereumSpecific.GetStatus` ?

platform/ethereum/blockbook/transaction.go
```diff
@@ -0,0 +1,146 @@
...
+	return ""
+}
+
+func getDirection(address, from, to string) blockatlas.Direction {
+	if address == from && address == to {
+		return blockatlas.DirectionSelf
+	}
+	if address == from {
+		return blockatlas.DirectionOutgoing
+	} else {
```
vikmeup, line 94: else not needed, just return

platform/ethereum/blockbook/transaction.go
```diff
@@ -0,0 +1,146 @@
...
+					metadata.From = transfer.From
+					metadata.To = address
+				}
+				final.Direction = direction
+				final.Meta = metadata
+				return
+			}
+		}
+	} else {
+		if tx.EthereumSpecific.GasLimit.Uint64() == 21000 {
```
vikmeup, line 133: transfer could be any gas limit - you should remove check for 21000

platform/ethereum/blockbook/block.go
```diff
@@ -0,0 +1,25 @@
...
+func (c *Client) GetBlockByNumber(num int64, coinIndex uint) (*blockatlas.Block, error) {
+	block, err := c.GetBlock(num)
+	if err != nil {
+		return nil, err
+	}
+
+	var txs []blockatlas.Tx
+	for _, srcTx := range block.Transactions {
+		tx := normalizeTx(&srcTx, coinIndex)
+		txs = append(txs, tx)
```
EnoRage, line 18: We need to filMeta for block txs also

platform/ethereum/blockbook/transaction.go
hewigovens: ideally I want to check tx data length, it's very hard to tell if a tx is a normal transfer or a contract call

platform/ethereum/blockbook/block.go
hewigovens: will do

platform/ethereum/blockbook/model.go
```diff
@@ -0,0 +1,77 @@
...
+type Output struct {
+	Value     string   `json:"value,omitempty"`
+	Addresses []string `json:"addresses"`
+}
+
+// TokenType specifies type of token
+type TokenType string
+
+// ERC20TokenType is Ethereum ERC20 token
+const ERC20TokenType TokenType = "ERC20"
```
vikmeup, line 48: can you reuse https://github.com/trustwallet/blockatlas/blob/3eeb1d248527e110a7b1a7ea08b965ff3c880b2e/pkg/blockatlas/tx.go#L18 ?

platform/ethereum/blockbook/transaction.go
hewigovens: done

platform/ethereum/blockbook/transaction.go
hewigovens: check GasUsed instead

platform/ethereum/blockbook/block.go
hewigovens: done

platform/ethereum/blockbook/model.go
hewigovens: done

platform/ethereum/base.go
```diff
@@ -3,30 +3,38 @@ package ethereum
...
 type Platform struct {
-	CoinIndex         uint
-	RpcURL            string
-	client            Client
-	collectionsClient CollectionsClient
-	ens               RpcClient
+	CoinIndex   uint
+	RpcURL      string
+	client      EthereumClient
+	tokens      EthereumClient
```
EnoRage, line 16: Do we use it somewhere?

EnoRage:
there is an issue with parsing transactions. Example:  block num 8698970

Trust RAY response:
Screenshot 2020-04-04 at 13 29 09

Blockbook:
Screenshot 2020-04-04 at 13 30 41

Transaction:
https://etherscan.io/tx/0x274df9abbb6009b4b08f22e44574e1b1a161eca0b01b7b77976774ab2f7a46b9

Another tx in this block with the same issue:
0x1d041cc3e10a71ef3cf6499c74fe03c1869a7c2b29e1dfed97316c2dcde63144


We need to ensure that TokenTransfer is not contract call type

EnoRage:
I think that the best way to check that: Get about 20 transactions from ray and ensure that blockbook will return the same response. We actively use Meta of tx to send notifications: https://github.com/trustwallet/blockatlas/blob/master/pkg/blockatlas/tx.go#L239
So tx.Meta must be set properly

hewigovens:
I'm fixing FIO issues, will test block against trust ray later

platform/ethereum/base.go
hewigovens: thanks, removed

hewigovens:
metadata missing in block api should be fixed, there are some differences from trust ray for token transfer

EnoRage:
1.
http://localhost:8420/v1/ethereum/0x11111254369792b2ca5d084ab5eea397ca8fa48b
Can we try to parse contract addresses? It is not working for ray implementation.

2. Also another exmaple with some strange symbols for USDC:
http://localhost:8420/v1/ethereum/0xde274664ad6ecaf0240501aff7d081fefc8654de
```
 {
            "id": "0x997b563a12fb10b2347783ee1c48f1557ea323179fd8d5bb34fdca5ea4c96d9d",
            "coin": 60,
            "from": "0xDe274664aD6ECaF0240501AFf7D081fEFC8654De",
            "to": "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48",
            "fee": "156180761014400",
            "date": 1585588433,
            "block": 9774314,
            "status": "completed",
            "sequence": 28,
            "type": "token_transfer",
            "direction": "outgoing",
            "memo": "",
            "metadata": {
                "name": "USD//C",
                "symbol": "USDC",
                "token_id": "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48",
                "decimals": 6,
                "value": "33507774130",
                "from": "0xDe274664aD6ECaF0240501AFf7D081fEFC8654De",
                "to": "0x69CEE776211DBC37Bf2a9f125f21670536f69359"
            }
        }
```

hewigovens:
1. didn't get it, we won't add any new features to trust ray
2. "USD//C" is from the blockbook api, we don't do any formating

EnoRage:
> 1. didn't get it, we won't add any new features to trust ray
> 2. "USD//C" is from the blockbook api, we don't do any formating

Currently there is 500 error for some addresses like 0x11111254369792b2ca5d084ab5eea397ca8fa48b
2. 👌

hewigovens:
I reproduced, because of pending txs, let me fix it

vikmeup:
> "USD//C" is from blockbook api, we don't do any formating

This is ok, as our backend will have correct name for it, we only need list of contracts

EnoRage:
Rancher was updated, necessary config endpoint provided

platform/ethereum/trustray/transaction.go
```diff
@@ -0,0 +1,123 @@
...
+
+func normalizePage(srcPage *Page, address string, coinIndex uint) blockatlas.TxPage {
+	var txs []blockatlas.Tx
+	for i, srcTx := range srcPage.Docs {
+		txs = AppendTxs(txs, &srcTx, coinIndex)
+		txs[i].Direction = txs[i].GetTransactionDirection(address)
+	}
+	page := blockatlas.TxPage(txs)
+	return page
+}
```
prazd, line 35: return blockatlas.TxPage(txs)

platform/ethereum/trustray/transaction.go
```diff
@@ -0,0 +1,123 @@
...
+			Symbol:   op.Contract.Symbol,
+			TokenID:  address.EIP55Checksum(op.Contract.Address),
+			Decimals: op.Contract.Decimals,
+			Value:    blockatlas.Amount(op.Value),
+			From:     op.From,
+			To:       op.To,
+		}
+		out = append(out, tokenTx)
+	}
+	return
```
prazd, line 84: I think you can delete this return

platform/ethereum/trustray/transaction.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+func extractBase(srcTx *Doc, coinIndex uint) (base blockatlas.Tx, ok bool) {
+	var status blockatlas.Status
+	var errReason string
```
prazd, line 89: var(
     status blockatlas.Status
     errReason string
)

hewigovens: thanks, all done, some are legacy code

platform/ethereum/collection.go
```diff
@@ -56,7 +58,7 @@ func NormalizeCollectiblePage(collectibles []Collectible, coinIndex uint) (page
 	return page
 }
 
-func NormalizeCollectible(a Collectible, coinIndex uint) blockatlas.Collectible {
+func NormalizeCollectible(a collection.Collectible, coinIndex uint) blockatlas.Collectible {
```
vikmeup, line 61: if collectible then `c`, not sure why `a`

platform/ethereum/collection/client.go
```diff
@@ -1,18 +1,19 @@
-package ethereum
+package collection
```
vikmeup, line 1: use package name as `collectibles` as `collection` as very generic word

platform/ethereum/trustray/client.go
```diff
@@ -0,0 +1,46 @@
...
+func (c *Client) GetCurrentBlockNumber() (int64, error) {
+	var nodeInfo NodeInfo
+	err := c.Get(&nodeInfo, "node_info", nil)
+	if err != nil {
+		return 0, err
+	}
+	return nodeInfo.LatestBlock, nil
+}
+
+func (c *Client) GetTokens(address string) (tp *TokenPage, err error) {
```
vikmeup, line 42: ideally this function just returns list of tokens, and not wrapped object with docs

platform/ethereum/base.go
```diff
@@ -3,30 +3,37 @@ package ethereum
...
 }
 
-func InitWitCollection(coin uint, api, rpc, collectionApi, collectionKey string) *Platform {
-	p := Init(coin, api, rpc)
-	p.collectionsClient = CollectionsClient{blockatlas.InitClient(collectionApi)}
-	p.collectionsClient.Headers["X-API-KEY"] = collectionKey
-	return p
+func InitWitCollection(coinType uint, api, rpc, blockbookApi, collectionApi, collectionKey string) *Platform {
+	platform := Init(coinType, api, rpc)
+	if coinType == coin.ETH {
```
vikmeup, line 31: can we improve this logic? blockbook only available on ETH node, why do we init collecitble here? if they are only ETH as well

platform/ethereum/blockbook/client.go
```diff
@@ -0,0 +1,55 @@
...
+	return nodeInfo.Blockbook.BestHeight, nil
+}
+
+func (c *Client) GetBlock(num int64) (block Block, err error) {
+	path := fmt.Sprintf("v2/block/%d", num)
+	err = c.Get(&block, path, nil)
+	return
+}
+
+func (c *Client) fetchTransactions(address, contract string) (page *Page, err error) {
```
vikmeup, line 41:
```suggestion
func (c *Client) getTransactions(address, contract string) (page *Page, err error) {
```

should we use Get() for consistency? same the function below

platform/ethereum/blockbook/client.go
```diff
@@ -0,0 +1,55 @@
...
+	path := fmt.Sprintf("v2/address/%s", address)
+	query := url.Values{"page": {"1"}, "pageSize": {"25"}, "details": {"txs"}, "contract": {contract}}
+	err = c.Get(&page, path, query)
+	return
+}
+
+func (c *Client) fetchTokens(address string) ([]Token, error) {
```
vikmeup, line 48: same here

platform/ethereum/blockbook/transaction.go
```diff
@@ -0,0 +1,139 @@
...
+				return
+			}
+		}
+	}
+	fillMeta(final, tx, coinIndex)
+}
+
+func fillMeta(final *blockatlas.Tx, tx *Transaction, coinIndex uint) {
+	gasUsed := tx.EthereumSpecific.GasUsed
+	if gasUsed != nil && gasUsed.Int64() == 21000 {
```
vikmeup, line 127: I think it's a bug - ETH transfers not always 21000 gas limit

platform/ethereum/blockbook/transaction.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+func normalizeBlockHeight(height int64) uint64 {
+	if height < 0 {
+		return uint64(0)
+	}
+	return uint64(height)
+}
+
+func FillMeta(final *blockatlas.Tx, tx *Transaction, coinIndex uint) {
```
vikmeup, line 71: `fillMeta` vs `FillMeta` little confusing naming on what it does?

platform/ethereum/collection.go
hewigovens: those are legacy code, will address them in a separate PR

platform/ethereum/blockbook/transaction.go
hewigovens: this checks `GasUsed` not (GasLimit), plain ether transfer always costs 21000 gas, we can improve this once we can check `len(data)`

platform/ethereum/trustray/client.go
hewigovens: this is also legacy code, we can improve it later

EnoRage:
@vikmeup are you ok with @hewigovens answers? If so @hewigovens let's merge it to master, Rancher is ready for that

vikmeup:
+

Issue #997, opened 2020-04-02
[PG] Duplicate key value violates unique constraint "subscriptions_pkey" 


vikmeup:
## Summary of Bug

```
(/build/db/subscriptions.go:56)
pq: duplicate key value violates unique constraint "subscriptions_pkey" 
ERRO[0229] pq: duplicate key value violates unique constraint "subscriptions_pkey"  id=2829854 operation=UpdateSubscription subscriptions_len=45
````

## Steps to Reproduce

____

##### Request details:
```
{}
```

##### Response details:
```
{}
```

PR #999, opened 2020-04-02
[Postgres fixes] and new field, add tests

EnoRage:
Add updated_at field, fix issue with conflct, add test. Closing #997 
Checklist: 
- [x] Fix issue
- [x] Add new field for TTL
- [x] Add tests
- [x] Test the migration
- [x] Add composite pk

https://gorm.io/docs/composite_primary_key.html#content-inner


db/subscriptions.go
```diff
@@ -136,7 +145,7 @@ func (i *Instance) BulkCreate(dataList []models.SubscriptionData) error {
 		valueArgs = append(valueArgs, d.Address)
 	}
 
-	smt := `INSERT INTO subscription_data(subscription_id, coin, address) VALUES %s`
+	smt := `INSERT INTO subscription_data(subscription_id, coin, address) VALUES %s ON CONFLICT DO NOTHING`
```
prazd, line 148: Maybe it's better to use constants for raw query?
What do you think?

PR #1001, opened 2020-04-03
[WIP] Adapt FIO lookup to mainnet FIO domains (@trust, ...).

catenocrypt:
So far only '@fiotestnet' was treated as FIO domain.  Include the following domains to be checked as a FIO name:

```
@trust
@trustwallet
@binance
@fiomembers
@  // any FIO domain
```

Note that any 'something@something' address will be assumed to be FIO, no other lookup scheme uses '@' as separator.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -150,7 +150,7 @@ nebulas:
 
 # [FIO] FIO: https://fioprotocol.io
 fio:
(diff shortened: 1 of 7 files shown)


hewigovens:
can you add FIO public key to test? coin 235

PR #1007, opened 2020-04-04
Return FIO transaction action sequence

hewigovens:


hewigovens:
found another issue, duplicated txs :(

Issue #1003, opened 2020-04-03
[Platform] Change Binance api from explorer to RPC 


EnoRage:
Explorer api is outdated, let's use RPC to prevent issues like constant fails of explorer

kolya182:
RPC not meant to serve as a source of transactions, it only keep history for 24 hours or so 
https://dex.binance.org/api/v1/transactions?address=bnb1xwalxpaes9r0z0fqdy70j3kz6aayetegur38gl&txType=TRANSFER&rows=25&page=1
vs
https://explorer.binance.org/api/v1/txs?address=bnb1pyzndgy0clkrehqke6h5jgrtdqzgdc78ac4qsr&page=1&rows=25

EnoRage:
Maybe we will need to write open source rpc?)

EnoRage:
https://dex.binance.org/api/v1/transactions?address=bnb1m7ly8xsty9rcfnhcdjnde4qtns27j0umpnv3ja

https://docs.binance.org/api-reference/dex-api/paths.html#apiv1transactions
max = 3 months

kolya182:
> https://dex.binance.org/api/v1/transactions?address=bnb1m7ly8xsty9rcfnhcdjnde4qtns27j0umpnv3ja
> 
> https://docs.binance.org/api-reference/dex-api/paths.html#apiv1transactions
> max = 3 months

This then
https://dex.binance.org/api/v1/transactions?address=bnb1pyzndgy0clkrehqke6h5jgrtdqzgdc78ac4qsr&txType=TRANSFER&rows=25&page=1&endTime=1578168727000

EnoRage:
Latest block
GetBlockByNumber
GetTokens of account

PR #1010, opened 2020-04-04
Migrate Binance platform to use only DEX Explorer API for transactions and RPC fo the rest calls

kolya182:
`Updated on May 8`
`Before merge`
Update env vars:
`binance.api` => `https://dex.binance.org/api`
`binance.dex` => `binance.explorer` => https://explorer.binance.org/api

Closes https://github.com/trustwallet/blockatlas/issues/1003

Removed code and tests related to unsupported by client type of transactions like `NEW_ORDER`, `CANCEL_ORDER`

TODO
- [x] Use RPC transactions hash endpoint to pull multi-send transfers, this fixed issue when address in multisend was not proceed as a result transfer was not returned.
- [x] Add test to model functions
- [x] Clean code

Fixed:
- [x] Missing transfers in multisend transaction, if requested address was deep in response that was not paginated by default, but assumed all multisend transfers returned in request to explorer.

As was tested RPC is not stable to return address transfers, and for now keep to use Explorer for `transactions only`
Example RPC: https://dex.binance.org/api/v1/transactions?address=bnb14cjy0yl23xkf0hnw3ql295v8nghqstvlzkvqpl&txType=TRANSFER
Example Explorer: https://explorer.binance.org/api/v1/txs?address=bnb14cjy0yl23xkf0hnw3ql295v8nghqstvlzkvqpl&page=1&rows=25&txType=TRANSFER


platform/binance/transaction.go
```diff
@@ -11,180 +12,111 @@ import (
...
+	tBase := blockatlas.Tx{
+		ID:        t.TxHash,
+		Coin:      coin.BNB,
+		From:      t.FromAddr,
+		To:        t.ToAddr,
+		Fee:       blockatlas.Amount(t.getFee()),
+		Date:      t.blockTimestamp(),
+		Block:     t.BlockHeight,
+		Status:    t.getStatus(),
+		Error:     "", // FIXME check for an error
```
EnoRage, line 68: Can you please finish that?

platform/binance/token.go
```diff
@@ -7,50 +7,48 @@ import (
...
+}
+
 // NormalizeToken converts a Binance token into the generic model
 func NormalizeToken(srcToken *Balance, tokens *TokenPage) (t blockatlas.Token, ok bool) {
 	if srcToken.isAllZeroBalance() {
 		return t, false
 	}
 
-	tk := tokens.findToken(srcToken.Symbol)
+	tk := tokens.findToken(srcToken.Symbol) // TODO create and use cache instead query tokens list every time
```
EnoRage, line 39: You can use GetWithCache to make HTTP GET request 
Example:
https://github.com/trustwallet/watchmarket/blob/master/market/clients/cmc/client.go#L49

EnoRage:
> Switching from explorer RPC to DEX RPC brings a few disadvantages:
> 
> No direct way to return address transactions including multi-send, but block transactions api supports multi-send (possible talks to Binance to include multi-send)

Is there any other issue? @kolya182 

Is it critical @vikmeup

vikmeup:
@EnoRage should be ok as long as we notify when receiving new transaction in the block and support for multi send will be added later for transactions API

platform/binance/client.go
```diff
@@ -16,36 +18,58 @@ type Client struct {
...
+// Get transactions in the block. Multi-send and multi-coin transactions are included as sub-transactions.
+func (c *Client) GetBlockTransactions(num int64) ([]TxV2, error) {
+	stx := new(BlockTxV2)
+	path := fmt.Sprintf("v2/transactions-in-block/%d", num)
+	err := c.Get(stx, path, nil)
+	return stx.Txs, err
+}
+
+//  Gets a list of address or token transactions by type
+func (c *Client) GetAddressAssetTrx(address, token, txType string) ([]Tx, error) {
```
vikmeup, line 37: do not use `Trx`. Use either `Transactions` or `Tx`, but more intuitive is: `GetAddressAssetTransactions`

platform/binance/model.go
```diff
@@ -31,103 +32,107 @@ type Error struct {
...
-		}
-		subTxValue, err := subTx.Value.Float64()
-		if err != nil {
-			subTxValue = 0
-		}
-		value := numbers.Float64toString(txValue + subTxValue)
-		tx.Value = json.Number(value)
-		mapTx[key] = tx
+type Token struct {
+	Mintable       bool   `json:"mintable"`
```
vikmeup, line 93: `Mintable` is not used

platform/binance/transaction.go
```diff
@@ -11,180 +12,111 @@ import (
...
-			// Recursive call to normalize the tx
-			newTxs, ok := normalizeTransfer(tx, subTx, token, address)
-			if !ok {
-				continue
-			}
-			txs = append(txs, newTxs...)
-		}
-		if len(txs) == 0 {
-			return txs, false
+// NormalizeTxs converts multiple Binance transactions
```
vikmeup, line 45:
```suggestion
// Converts multiple transactions
```

platform/binance/transaction.go
```diff
@@ -11,180 +12,111 @@ import (
...
+		if !ok {
+			continue
 		}
-		return txs, true
+		txs = append(txs, tx...)
+	}
+	return
+}
+
+// NormalizeTx converts a Binance transaction into the generic model
```
vikmeup, line 57:
```suggestion
// Converts a transaction into the generic model
```

platform/binance/model.go
```diff
@@ -31,103 +32,107 @@ type Error struct {
...
+	Fee      string `json:"txFee"`
+	FromAddr string `json:"fromAddr"`
+	Hash     string `json:"txHash"`
+	ToAddr   string `json:"toAddr"`
+	Type     TxType `json:"txType"`
+	Value    string `json:"value"`
 }
 
-type SubTxs []SubTx
+type TokenPage []Token
```
vikmeup, line 90:
```suggestion
type TokenList []Token
```

platform/binance/model.go
```diff
@@ -31,103 +32,107 @@ type Error struct {
...
+	Sequence    uint64 `json:"sequence"`
+	Source      int    `json:"source"`
+	Timestamp   string `json:"timeStamp"`
+	ToAddr      string `json:"toAddr"`
+	TxHash      string `json:"txHash"`
+	Type        TxType `json:"txType"`
+	Value       string `json:"value"`
+}
+
+type BlockTxV2 struct {
```
vikmeup, line 68:
```suggestion
type BlockTransactions struct {
```

platform/binance/model.go
```diff
@@ -31,103 +32,107 @@ type Error struct {
...
 }
 
 type TxType string
 
-const (
-	TxTransfer    TxType = "TRANSFER"
-	TxNewOrder    TxType = "NEW_ORDER"
-	TxCancelOrder TxType = "CANCEL_ORDER"
-)
+type TransactionsV1 struct {
```
vikmeup, line 45:
```suggestion
type Transactions struct {
```

platform/binance/block.go
```diff
@@ -2,36 +2,49 @@ package binance
...
-	var txs blockatlas.TxPage
-	childTxs, err := p.getTxChildChan(srcTxs.Txs)
-	if err == nil {
-		txs = NormalizeTxs(childTxs, "", "")
-	} else {
-		txs = NormalizeTxs(srcTxs.Txs, "", "")
+	childTxs := make([]Tx, 0)
+	for _, t := range blockTxs {
+		if len(t.SubTransactions) > 0 {
+			for _, tSub := range t.SubTransactions {
```
vikmeup, line 25: add separate method to normalize

platform/binance/token.go
vikmeup: @kolya182 remove TODO if you added cache

platform/binance/client.go
```diff
@@ -16,36 +18,58 @@ type Client struct {
...
+	stx := new(BlockTransactions)
+	path := fmt.Sprintf("v2/transactions-in-block/%d", num)
+	err := c.Get(stx, path, nil)
+	return stx.Txs, err
+}
+
+//  Gets a list of address or token transactions by type
+func (c *Client) GetAddressAssetTransactions(address, token, txType string) ([]Tx, error) {
+	if address == "" && token == "" {
+		return []Tx{}, errors.E("Address and token not specified")
```
prazd, line 39: return nil, ...

platform/binance/client.go
```diff
@@ -16,36 +18,58 @@ type Client struct {
...
+	return account, err
 }
 
-func (c *Client) GetTx(hash string) (stx Tx, err error) {
-	err = c.Get(&stx, "tx", url.Values{"txHash": {hash}})
-	return
+// Gets a list of tokens that have been issued.
+func (c *Client) GetTokens() (*TokenList, error) {
+	stp := new(TokenList)
+	query := url.Values{"limit": {"1000"}}
```
prazd, line 70: Maybe use consant for limit?

EnoRage:
NRY

platform/binance/client.go
```diff
@@ -2,83 +2,129 @@ package binance
...
+	err := c.Get(result, "v1/node-info", nil)
 	return result, err
 }
 
-func (c *Client) GetBlockByNumber(num int64) (*TxPage, error) {
-	stx := new(TxPage)
+// Get transactions in the block. Multi-send and multi-coin transactions are included as sub-transactions.
+func (c *Client) GetBlockTransactions(num int64) ([]TxV2, error) {
+	stx := new(BlockTransactions)
+	path := fmt.Sprintf("v2/transactions-in-block/%d", num)
```
EnoRage, line 34: http://dex.binance.org/api/v2/transactions-in-block/80157267?
Not working now

platform/binance/client.go
```diff
@@ -16,10 +16,15 @@ type Client struct {
 	blockatlas.Request
 }
 
+const (
+	TokensLimit       = "1000"
+	TransactionsLimit = "25"
```
vikmeup, line 21: `"25"` => `25`

make this a global parameter, as we always return the last 25

kolya182: https://github.com/trustwallet/blockatlas/pull/1010/commits/7559cf323139c706b8e44cfb911e7eed2f5106d6

platform/binance/client.go
```diff
@@ -18,7 +18,6 @@ type Client struct {
 
 const (
 	TokensLimit       = "1000"
```
vikmeup, line 20: same for `"1000"`

platform/binance/client.go
```diff
@@ -82,7 +82,7 @@ func (c *Client) GetAccountMetadata(address string) (*Account, error) {
 // Gets a list of tokens that have been issued.
 func (c *Client) GetTokens() (*TokenList, error) {
 	var (
-		query = url.Values{"limit": {TokensLimit}}
+		query = url.Values{"limit": {string(TokensLimit)}}
```
vikmeup, line 85: proper way to use `strconv.Itoa()`

platform/binance/client.go
erhenglu: Could you please use rpc tx_search instead of use this API? This API is not rpc, still costing a lot of resource 

curl "localhost:27147/tx_search?query=\"tx.height=10000\"&prove=true"
https://github.com/binance-chain/java-sdk/wiki/RPC#query-transactions-by-block
https://docs.binance.org/api-reference/node-rpc.html#5-rpc-endpoint-list

EnoRage:
@kolya182 
Find an issue with v2/tokens

http://localhost:8420/v2/binance/tokens/bnb1sj4hr4ll94xsmpuf9s296ufszahg0gsxdjn7d5
https://explorer.binance.org/address/bnb1sj4hr4ll94xsmpuf9s296ufszahg0gsxdjn7d5

http://localhost:8420/v2/binance/tokens/bnb10uetf9rxqxxkc67xf6799pl82gdd2gshwcqnuy
https://explorer.binance.org/address/bnb10uetf9rxqxxkc67xf6799pl82gdd2gshwcqnuy

More examples:
/v2/binance/tokens/bnb1y2akqpljrxmnxmz4qx2s8p7fnvh3euput504w0
/v2/binance/tokens/bnb17h49lval8vu23hrrmjjg64tdwcnl03v4tqepcf
/v2/binance/tokens/bnb1pzs4ugyuhtdg00fpj2waqtrppwa7wm7h83lgzk
/v2/binance/tokens/bnb1ujtqf69mctgcan3l3fgw5pgtxy7a9lqlep6cx8

platform/binance/block.go
```diff
@@ -2,38 +2,53 @@ package binance
...
+			}
+		}
+	}
+
+	txs := NormalizeTxs(childTxs, "")
+
+	return &blockatlas.Block{Number: num, Txs: txs}, nil
+}
+
+func normalizeBlockSubTx(t *TxV2, tSub *SubTx) Tx {
```
prazd, line 36: Maybe normalizeBlockSubTx(t TxV2, tSub SubTx)?
Because you don't change t and tSub in this func

platform/binance/model.go
```diff
@@ -31,101 +32,104 @@ type Error struct {
...
+func (tx *Tx) getStatus() blockatlas.Status {
+	switch tx.Code {
+	case 0:
+		return blockatlas.StatusCompleted
+	default:
+		return blockatlas.StatusError
+	}
+}
+
+func (tx *Tx) getError() string {
```
prazd, line 117: Maybe return bool value?
Like that:

```
func (tx *Tx) isCompleted() bool {
switch tx.getStatus() {
	case blockatlas.StatusCompleted:
		return true
	default:
		return false
    }
}
```

platform/binance/transaction.go
```diff
@@ -11,180 +12,111 @@ import (
...
+	// Condition for BEP2 token transfer e.g.: TWT-8C2
+	t.Type = blockatlas.TxNativeTokenTransfer
+	t.Meta = blockatlas.NativeTokenTransfer{
 		Decimals: bnbCoin.Decimals,
 		From:     srcTx.FromAddr,
+		Symbol:   TokenSymbol(srcTx.Asset),
 		To:       srcTx.ToAddr,
+		TokenID:  srcTx.Asset,
+		Value:    value,
+		Name:     "",
```
prazd, line 110: if I am not mistaken, this can be skipped (the default value is "")

EnoRage:
@catenocrypt can you please help to fix mocked test? I cannot understand why do we need 
	@$(MAKE) newman-run test=mock-healthcheck host=http://localhost:8437
and empty files

pkg/numbers/number.go
```diff
@@ -32,10 +32,21 @@ func Float64toPrecision(num float64, precision int) float64 {
...
 }
 
+// 0.1010 => "0.101"
 func Float64toString(num float64) string {
 	return strconv.FormatFloat(num, 'f', -1, 64)
 }
 
+// "0.00037500" => 0.000375
+func StringNumberToFloat64(str string) float64 {
+	value, err := strconv.ParseFloat(str, 64)
```
vikmeup, line 42: return an error if failed

pkg/numbers/number_test.go
```diff
@@ -64,4 +64,10 @@ func TestFloat64toString(t *testing.T) {
 	assert.Equal(t, Float64toString(1), "1")
 	assert.Equal(t, Float64toString(1.1), "1.1")
 	assert.Equal(t, Float64toString(1.015), "1.015")
+	assert.Equal(t, Float64toString(2800.00000000), "2800")
+}
+
+func TestStringNumberToFloat64(t *testing.T) {
+	assert.Equal(t, 0.2997, StringNumberToFloat64("0.29970000"))
```
vikmeup, line 71: add fail case

platform/binance/block.go
```diff
@@ -18,37 +20,67 @@ func (p *Platform) GetBlockByNumber(num int64) (*blockatlas.Block, error) {
...
+
+	time, err := time.Parse(time.RFC3339, t.Timestamp)
+	if err != nil {
+		tx.Timestamp = time.Unix()
+	}
+
+	var multisend []multiTransfer
+	for _, st := range t.SubTransactions {
+		v := multiTransfer{
+			Amount: numbers.ToDecimal(st.Value, 8),
```
vikmeup, line 76: Do not use `8`, use `coin.Binance().decimals` variable

platform/binance/transaction.go
kolya182: Yes and know, ideally we need to return token name for as far is it defined on struct. For now removed.

platform/binance/block.go
kolya182: Bit modified this function, please take a look.

platform/binance/model.go
kolya182: Error is type `string`, i is it make sense to return string as is it now.
Screen Shot 2020-05-03 at 10 18 32 PM

platform/binance/block.go
```diff
@@ -2,38 +2,93 @@ package binance
...
+			continue
+		}
+		normTxs = append(normTxs, normT...)
+	}
+
+	return &blockatlas.Block{Number: num, Txs: normTxs}, nil
+}
+
+// Normalize block sub transaction from RPC to explorer transaction
+func normalizeBlockSubTx(t *TxV2) DexTx {
```
prazd, line 42: Also you don't change TxV2 in this func
You create a `tx` object and modify
Imho `normalizeBlockSubTx(txV2 TxV2) DexTx {` or `normalizeBlockSubTx(txV2 TxV2) *DexTx {` 
will be better
What do you think?

catenocrypt:
Usage of API, DEX, and RPC should be clarified, it's a total mess now.  3 names used for 2 things, in conflicting way:
```
		rpcClient: Client{blockatlas.InitClient(api)},
		dexClient: DexClient{blockatlas.InitClient(dex)},
```
Here RPC goes with API

```
  api: https://dex.binance.org/api
  dex: https://explorer.binance.org/api
```
Here dex goes with api.

platform/binance/block.go
```diff
@@ -2,38 +2,93 @@ package binance
...
-		Txs:    txs,
-	}, nil
+
+	time, err := time.Parse(time.RFC3339, t.Timestamp)
+	if err != nil {
+		tx.Timestamp = time.Unix()
+	}
+
+	var multisend []multiTransfer
+	for _, st := range t.SubTransactions {
```
prazd, line 81:
```
multisend := make([]multiTransfer, 0, len(t.SubTransactions))
```

config.yml
```diff
@@ -39,11 +39,10 @@ observer:
...
   uri: postgresql://user:pass@localhost/my_db?sslmode=disable
 
-# [BNB] Binance DEX: https://wallet.binance.org
-#       Binance Chain: https://explorer.binance.org
+# [BNB] Binance DEX: https://www.binance.org/
 binance:
-  api: https://explorer.binance.org/api/v1
-  dex: https://dex.binance.org/api
+  api: https://dex.binance.org/api
+  dex: https://explorer.binance.org/api/v1
```
catenocrypt, line 45: Move the 'v1' version designator from config to client code -- versioning is not a configuration issue, but a code issue.  If version is changed, likely code has to be changed as well.
Remove v1 from config, and add it to platform/binance/dex_client.go .

configmock.yml
```diff
@@ -37,8 +37,8 @@ postgres:
 # [BNB] Binance DEX: https://wallet.binance.org
 #       Binance Chain: https://explorer.binance.org
 binance:
-  api: http://localhost:3347/binance-api/v1
-  dex: http://localhost:3347/binance-dex
+  api: https://dex.binance.org/api
+  dex: https://localhost:3347/api/v1
```
catenocrypt, line 41: localhost:3347 is the mock API server, for **ALL** mock APIs, therefore mock API paths have to be prefixed with the platform, like this:  http://localhost:3347/binance-rpc  (or -dex).  Just using 'localhost:3347/api/v1' for a mock path is wrong.

Note: mock script contain there path in the 2nd line usually:
module.exports = {  path: '/binance-rpc/...

Note: When `start-platform-api-mock` is started, it displays diagnostic output on the console about mock API calls received.

mock/ext-api-dyson/get/binance-rpc-txs.js
```diff
@@ -0,0 +1,69 @@
...
+/// Returns:
+/// - Multi-transaction transaction for a specific address
+///   1. newman calls: http://localhost:8437/v1/binance/bnb1jeu6gscugy6l2wyatxthkh2hmer4hzevgcmf0q
+///   2. Block atlas call internally : https://{binance_rpc}/api/v1/txs?address=bnb1jeu6gscugy6l2wyatxthkh2hmer4hzevgcmf0q&page=1&rows=25&txType=TRANSFER
+///   3. Dyson response mock data: http://localhost:3347/binance-dex/api/v1/txs?address=bnb1jeu6gscugy6l2wyatxthkh2hmer4hzevgcmf0q&page=1&rows=25&txType=TRANSFER
+/// - empty response for other txHash'es
+
+// Example contains each type, BNB transfer, BEP2 transfer, BEP2 transfer as multisend transaction
+module.exports = {
+    path: '/binance-dex/v1/transactions',
```
catenocrypt, line 11: The path here does not match the paths used to DEX, shown in lines 2. and 3. above.  I think the correct would be '/binance-dex/v1/txs'.

kolya182:
@prazd @catenocrypt I made changes based on your suggestions.
@catenocrypt I agree that naming  `rpc` `api` `dex` is confusing. I made simple change for binance platform for `dex` => `explorer` (not touching api which is should be common for all platforms in current implementation, other way it will not start). Lets discuss how we can improve it in next PR's

platform/binance/transaction.go
```diff
@@ -1,190 +1,167 @@
...
+		wg      sync.WaitGroup
+		hashChan = make(chan TxHashRPC)
+	)
+
+	for i, srcT := range srcTxs {
+		if srcT.HasChildren == 1 {
+			wg.Add(1)
+			go func(srcTx DexTx, out chan TxHashRPC, wg *sync.WaitGroup) {
+				defer wg.Done()
+				defer close(out)
```
prazd, line 50: If I'm not mistaken, panic should happen here, because you close the channel in one procedure, but in other programs you continue to write to it.
Solution - close the channel after `wg.Wait()`
It is also worth testing this func (with `-race` flag too)

EnoRage:
It is not ready. 

We need to change normalization methods. It is not a good way to:
1) Fetch data to struct A
2) Modify struct A 
3) Normalize it to struct B
4) Normalize it to struct C

It is important to keep as much as possible normalizations.

We need to cover all the client methods as well as normalization methods for different cases + negative cases.

We already found some errors that were not tracked by tests at all or test was incorrect (excepted data was wrong)

I will start to refactor it. I will need some help with tests

platform/binance/transaction.go
```diff
@@ -71,7 +89,7 @@ func normalizeSingleTransfer(srcTx ExplorerTxs, address, token string) blockatla
 		return blockatlas.TxPage{tx}
 	}
 
-	if srcTx.TxAsset == token {
+	if srcTx.TxAsset != "" {
```
kolya182, line 92: toke can be anything in this case, not exact match

EnoRage: what do you mean?

EnoRage:
@kolya182 can you please cover it with tests and finish before making another commits to atlas, please?

EnoRage:
Screenshot 2020-05-14 at 22 49 45
parser issue with metadata

kolya182:
> Screenshot 2020-05-14 at 22 49 45
> parser issue with metadata

Can you indignity exact block where it happened ?

platform/binance/transaction.go
kolya182: Your changes introduced serious bug:
`curl --location --request GET 'http://localhost:8420/v1/binance/bnb14cjy0yl23xkf0hnw3ql295v8nghqstvlzkvqpl'`
return BEP2 transfers also.

kolya182:
> It is not ready.
> 
> We need to change normalization methods. It is not a good way to:
> 
> 1. Fetch data to struct A
> 2. Modify struct A
> 3. Normalize it to struct B
> 4. Normalize it to struct C
> 
> It is important to keep as much as possible normalizations.
> 
> We need to cover all the client methods as well as normalization methods for different cases + negative cases.
> 
> We already found some errors that were not tracked by tests at all or test was incorrect (excepted data was wrong)
> 
> I will start to refactor it. I will need some help with tests

Can you be more specific  on `some errors` ?

platform/binance/model.go
```diff
@@ -266,6 +266,10 @@ func (e *Error) Error() string {
 
 // Determines Explorer transaction direction relatively to address
 func (tx *ExplorerTxs) Direction(address string) blockatlas.Direction {
+	if address == "" {
+		return ""
```
vikmeup, line 270: can you return `blockatlas.Direction` type here?

EnoRage: It is blockatls.Direction type. Golang do cast instead of me automatically. Because return type is blockatlas.Direction is string

platform/binance/block.go
```diff
@@ -2,38 +2,74 @@ package binance
...
+		tx.TxFee = fee
+	}
+	if len(txV2.SubTransactions) > 0 {
+		tx.HasChildren = 1
+	}
+	if t, err := time.Parse(time.RFC3339, txV2.Timestamp); err == nil {
+		tx.Timestamp = t.Unix()
+	}
+
+	mts := make([]MultiTransfer, len(txV2.SubTransactions))
```
prazd, line 69: nice!

platform/binance/explorer_client.go
```diff
@@ -0,0 +1,25 @@
...
+	blockatlas.Request
+}
+
+const (
+	explorerRows = "25"
+	explorerPage = "1"
+)
+
+func (c ExplorerClient) getTxsOfAddress(address, token string) (ExplorerResponse, error) {
+	stx := new(ExplorerResponse)
```
prazd, line 18: change naming

services/observer/subscriber/subscriber_test.go
```diff
@@ -0,0 +1,10 @@
+package subscriber
+
+import (
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func TestSubscriberVariables(t *testing.T) {
+	assert.Equal(t, AddSubscription, AddSubscription)
+}
```
prazd, line 10: ?

PR #1012, opened 2020-04-05
Change the pipeline: add golangcli-lint, fix errors with it

EnoRage:


platform/aeternity/client.go
```diff
@@ -11,15 +11,15 @@ type Client struct {
...
 		"limit": {strconv.Itoa(limit)},
 	}
 	uri := fmt.Sprintf("middleware/transactions/account/%s", address)
-
-	err = c.Get(&transactions, uri, query)
+	var transactions []Transaction
+	err := c.Get(&transactions, uri, query)
 	if err != nil {
-		return
+		return nil, err
```
prazd, line 22: if err := c.Get(...); err != nil {
...

Issue #1018, opened 2020-04-08
[Platform-API] Reverse name lookup


catenocrypt:
(Rationale) It would be nice to display human-friendly names instead of raw addresses in transaction history and other places in Trust Wallet.  For this, a reverse name lookup is needed (return name from raw address, if available).

To check: Which naming systems (ENS, unstoppable, FIO, ...) support reverse lookup (if any).

Solution:  
- New blockatlas API for reverse name lookup
- For naming systems with reverse lookup possibility, forward call and also use a cache.
- For naming systems with no reverse lookup possibility: maintain cache of name-address associations; return answer from the cache.  Update cache whenever a forward name lookup is performed (asynchronously).  Note that in case of cache miss there is no way to trigger search, so functionality will work only for addresses already seen by blockatlas; also results may be outdated (minor issue).

EnoRage:
It is good idea just to add another field for "from", "to" addresses.
So it will be availible to return transactions like

"id": ..
"from": "0x...."
"from_domains":["vikmeup.eth," "viktor.eth", ...],
"to":"0x...",
"to_domains":["exchange.eth"],

EnoRage:
The only issue is the perfomance. I don’t think that it is possible without caching to do that for transactions

hewigovens:
I'm going to implement ENS/FIO reverse lookup, so it's easier for app to show name associated with their address or public key

vikmeup:
Let's close this for now, not needed atm

PR #1027, opened 2020-04-10
Implement reverse lookup for ENS/FIO

hewigovens:
part of #1018 

- [ ] mock tests

ENS:

`/ns/reverse?coin=60&address=0xC36edF48e21cf395B206352A1819DE658fD7f988`

```json
[
  {
    "result": "ohmyname.eth",
    "coin": 60
  }
]
```

FIO:

`ns/reverse?coin=235&address=FIO6XKiobBGrmPZDaHne47TF25CamjFc3cRe8hs7oAiCN59TbJzes`

```json
[
  {
    "result": "ip@trust",
    "coin": 235
  },
  {
    "result": "pimptee@trust",
    "coin": 235
  }
]
```


catenocrypt:
Excellent!

Caching will be needed, because this may be a high-traffic API.  Most of the request will be negative (no name found), I think those should be cached as well (with a rel. short expiry).

Mocked tests I can add (after this PR).

One issue: FIO can potentially handle any kind of address, but I think in this impl. only FIO addresses will be attempted (NamingAPIs map).  If so, this should be changed.

hewigovens:
This is mainly for receiving screen in apps, you share your ETH/FIO address

catenocrypt:
> This is mainly for receiving screen in apps, you share your ETH/FIO address

Also for transaction history view.

hewigovens:
Will reopen it when app really needs it

PR #1039, opened 2020-04-20
Exclude mocked js from repo language statistics

hewigovens:
Although it's not vendor code, I think it's the easiest way to exclude it

From https://github.com/github/linguist#vendored-code


hewigovens:
It should be a golang project image

catenocrypt:
👍   Go: up Js: down

PR #1045, opened 2020-04-22
Do not return ERC20 tokens with zero balance

kolya182:
* Flatten test data and add test case fro modified logic. Modify test to increase test coverage
* Uncomment and add public endpoints to Ethereum config

`Note`: Currently only platform with `api` setting set up eg (ethereum.api), but since we moved away from trust-ray for ETH (api was representation of it) i have uncommented with url to localhost in order being able to setup platform. 

- [ ] Fix mock test

### DIFF
--- a/platform/ethereum/blockbook/model.go
+++ b/platform/ethereum/blockbook/model.go
@@ -58,7 +58,7 @@ type TokenTransfer struct {
 
 // Token contains info about tokens held by an address
 type Token struct {
-	Balance  string               `json:"balance"`
+	Balance  string               `json:"balance,omitempty"`
 	Contract string               `json:"contract"`
 	Decimals uint                 `json:"decimals"`
 	Name     string               `json:"name"`
--- a/platform/ethereum/blockbook/model.go
+++ b/platform/ethereum/blockbook/model.go
@@ -58,6 +58,7 @@ type TokenTransfer struct {
 
 // Token contains info about tokens held by an address
 type Token struct {
+	Balance  string               `json:"balance"`
 	Contract string               `json:"contract"`
 	Decimals uint                 `json:"decimals"`
 	Name     string               `json:"name"`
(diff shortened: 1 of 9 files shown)


platform/ethereum/blockbook/model.go
```diff
@@ -58,6 +58,7 @@ type TokenTransfer struct {
 
 // Token contains info about tokens held by an address
 type Token struct {
+	Balance  string               `json:"balance"`
```
hewigovens, line 61:
```suggestion
	Balance  string               `json:"balance,omitempty"`
```
this might be missing empty in some apis

platform/ethereum/blockbook/token.go
```diff
@@ -16,6 +16,9 @@ func (c *Client) GetTokenList(address string, coinIndex uint) (blockatlas.TokenP
 func NormalizeTokens(srcTokens []Token, coinIndex uint) []blockatlas.Token {
 	tokenPage := make([]blockatlas.Token, 0, len(srcTokens))
 	for _, srcToken := range srcTokens {
+		if srcToken.Balance == "0" {
```
hewigovens, line 19: also check empty string

platform/ethereum/blockbook/token_test.go
```diff
@@ -9,39 +9,27 @@ import (
...
-			want: blockatlas.Token{
-				Name:     "USD//C",
-				Symbol:   "USDC",
-				Decimals: 6,
-				TokenID:  "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48",
-				Type:     "ERC20",
-				Coin:     60,
-			},
+			name: "Should normalize and return token with balance",
+			args: args{srcToken: Token{Balance: "100", Type: "ERC20", Name: "USD//C", Contract: "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48", Symbol: "USDC", Decimals: 6}, coinIndex: 60},
```
hewigovens, line 22: a bit hard to read, can we keep old format?

PR #1051, opened 2020-04-24
Fix doge inputs value

### LINKED ISSUE #981: [Platform] Dogecoin Transaction Inputs
Blockatlas has a bug when you receive Dogecoin transactions from it.
If you look at the "inputs", you will see the numbers doesn't have the decimal places and is incorrect.

Here is an example:

Dogechain API:
https://dogechain.info/api/v1/transaction/b02977b96e5c65fd807e28230375c1267ded1de7c2c43292bf36552283bc5696

Inputs sum: 10

Blockatlas API:

http://localhost:8420/v1/doge/address/DPoYGk1wGQ3uWs5G3exd9WKvVyu8weKYVA
image

Inputs sum: 10
But if you look at the other values, they have the decimal places, just this value not.

hewigovens:
Fixes #981

### DIFF
--- a/.gitattribute
+++ b/.gitattribute
@@ -1,0 +1,1 @@
+mock/ext-api-dyson/* linguist-vendored
--- a/coin/coins.go
+++ b/coin/coins.go
@@ -454,7 +454,7 @@ var Coins = map[uint]Coin{
 		Decimals:         8,
 		BlockTime:        60000,
 		MinConfirmations: 0,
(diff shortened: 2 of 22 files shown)


kolya182:
@hewigovens I also was not able to alight mock json, brutal 😂

kolya182:
So v2 is giving just correct data vs v1 ?. Small comment would be appreciated.

hewigovens:
@kolya182 xpub api is v2 based, should work fine

PR #1056, opened 2020-04-25
[API] Add GetTokens handler

prazd:
Add new handler
Now we can get list of tokens by map: coin -> [addresses]

POST
path - `/v3/tokens`
body - `{"coin": ["address"]}`

### DIFF
--- a/api/endpoint/token.go
+++ b/api/endpoint/token.go
@@ -43,8 +43,8 @@ func GetTokensByAddress(c *gin.Context, tokenAPI blockatlas.TokenAPI) {
 // @Tags Transactions
 // @Param data body string true "Payload" default({"60": ["0xb3624367b1ab37daef42e1a3a2ced012359659b0"]})
 // @Success 200 {object} blockatlas.ResultsResponse
-// @Router /v3/tokens [post]
-func GetTokens(c *gin.Context, tokensAPIs map[uint]blockatlas.TokenAPI) {
+// @Router /v2/tokens [post]
+func GetTokens(c *gin.Context, apis map[uint]blockatlas.TokenAPI) {
 	var query map[string][]string
 	if err := c.Bind(&query); err != nil {
 		c.JSON(http.StatusInternalServerError, model.CreateErrorResponse(model.InternalFail, err))
@@ -56,12 +56,12 @@ func GetTokens(c *gin.Context, tokensAPIs map[uint]blockatlas.TokenAPI) {
 		if err != nil {
 			continue
 		}
-		tokenAPI, ok := tokensAPIs[uint(coinNum)]
+		api, ok := apis[uint(coinNum)]
 		if !ok {
 			continue
 		}
 
-		tokens := getTokens(tokenAPI, addresses)
+		tokens := getTokens(api, addresses)
 		result = append(result, tokens...)
 	}
 	c.JSON(http.StatusOK, blockatlas.ResultsResponse{Total: len(result), Results: &result})
--- a/api/endpoint/token.go
+++ b/api/endpoint/token.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"strconv"
 	"sync"
+	"time"
 )
 
 // @Summary Get Tokens
@@ -79,12 +80,25 @@ func getTokens(tokenAPI blockatlas.TokenAPI, addresses []string) blockatlas.Toke
 		go func(address string, wg *sync.WaitGroup) {
 			defer wg.Done()
 
-			tokenPage, err := tokenAPI.GetTokenListByAddress(address)
-			if err != nil {
-				return
-			}
+			stopChan := make(chan struct{})
+			pageChan := make(chan blockatlas.TokenPage)
 
-			tokenPagesChan <- tokenPage
+			go func() {
+				tokenPage, err := tokenAPI.GetTokenListByAddress(address)
+				if err != nil {
+					stopChan <- struct{}{}
+				}
+				pageChan <- tokenPage
+			}()
+
+			select {
+			case <-time.After(time.Second * 3):
+				return
+			case <-stopChan:
+				return
+			case p := <-pageChan:
+				tokenPagesChan <- p
+			}
 		}(address, &wg)
 	}
 	wg.Wait()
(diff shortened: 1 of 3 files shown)


api/endpoint/token.go
```diff
@@ -32,3 +34,65 @@ func GetTokensByAddress(c *gin.Context, tokenAPI blockatlas.TokenAPI) {
...
 	c.JSON(http.StatusOK, blockatlas.DocsResponse{Docs: &result})
 }
+
+// @Description Get tokens
+// @ID tokens_v3
+// @Summary Get list of tokens by map: coin -> [addresses]
+// @Accept json
+// @Produce json
+// @Tags Transactions
+// @Param data body string true "Payload" default({"60": ["0xb3624367b1ab37daef42e1a3a2ced012359659b0"]})
```
kolya182, line 44: What is the purpose of default ?

api/registry.go
```diff
@@ -90,6 +90,9 @@ func RegisterBatchAPI(root gin.IRouter) {
 	root.POST("/v4/collectibles/categories", func(c *gin.Context) {
 		endpoint.GetCollectionCategoriesFromList(c, platform.CollectionAPIs)
 	})
+	root.POST("/v3/tokens", func(c *gin.Context) {
```
vikmeup, line 93:
```suggestion
	root.POST("/v1/tokens", func(c *gin.Context) {
```

did we use v1 and v2 alredy?

api/endpoint/token.go
```diff
@@ -32,3 +34,65 @@ func GetTokensByAddress(c *gin.Context, tokenAPI blockatlas.TokenAPI) {
...
+		c.JSON(http.StatusInternalServerError, model.CreateErrorResponse(model.InternalFail, err))
+		return
+	}
+	result := make(blockatlas.TokenPage, 0)
+	for coinStr, addresses := range query {
+		coinNum, err := strconv.ParseUint(coinStr, 10, 32)
+		if err != nil {
+			continue
+		}
+		tokenAPI, ok := tokensAPIs[uint(coinNum)]
```
vikmeup, line 59:
```suggestion
		api, ok := tokensAPIs[uint(coinNum)]
```

api/endpoint/token.go
```diff
@@ -32,3 +34,65 @@ func GetTokensByAddress(c *gin.Context, tokenAPI blockatlas.TokenAPI) {
...
+		if !ok {
+			continue
+		}
+
+		tokens := getTokens(tokenAPI, addresses)
```
vikmeup, line 64:
```suggestion
		tokens := getTokens(api, addresses)
```

api/endpoint/token.go
```diff
@@ -32,3 +34,65 @@ func GetTokensByAddress(c *gin.Context, tokenAPI blockatlas.TokenAPI) {
...
+// @Router /v3/tokens [post]
+func GetTokens(c *gin.Context, tokensAPIs map[uint]blockatlas.TokenAPI) {
+	var query map[string][]string
+	if err := c.Bind(&query); err != nil {
+		c.JSON(http.StatusInternalServerError, model.CreateErrorResponse(model.InternalFail, err))
+		return
+	}
+	result := make(blockatlas.TokenPage, 0)
+	for coinStr, addresses := range query {
+		coinNum, err := strconv.ParseUint(coinStr, 10, 32)
```
vikmeup, line 55: line 55 should already return `uint` value ?

api/endpoint/token.go
```diff
@@ -32,3 +34,65 @@ func GetTokensByAddress(c *gin.Context, tokenAPI blockatlas.TokenAPI) {
...
+// @Description Get tokens
+// @ID tokens_v3
+// @Summary Get list of tokens by map: coin -> [addresses]
+// @Accept json
+// @Produce json
+// @Tags Transactions
+// @Param data body string true "Payload" default({"60": ["0xb3624367b1ab37daef42e1a3a2ced012359659b0"]})
+// @Success 200 {object} blockatlas.ResultsResponse
+// @Router /v3/tokens [post]
+func GetTokens(c *gin.Context, tokensAPIs map[uint]blockatlas.TokenAPI) {
```
vikmeup, line 47:
```suggestion
func GetTokens(c *gin.Context, apis map[uint]blockatlas.TokenAPI) {
```

api/endpoint/token.go
```diff
@@ -32,3 +34,65 @@ func GetTokensByAddress(c *gin.Context, tokenAPI blockatlas.TokenAPI) {
...
+		result         blockatlas.TokenPage
+	)
+
+	for _, address := range addresses {
+		wg.Add(1)
+		go func(address string, wg *sync.WaitGroup) {
+			defer wg.Done()
+
+			tokenPage, err := tokenAPI.GetTokenListByAddress(address)
+			if err != nil {
```
vikmeup, line 83: can you set timeout of **3 seconds** timeout? otherwise return results that were completed

api/endpoint/token.go
prazd: just body example for swagger

api/endpoint/token.go
prazd: `uint64`
but coin in map is `uint`

api/registry.go
prazd: I think we can use `v2` here too

api/endpoint/token.go
prazd: I'll think about it

prazd:
> We need new architecture for all atlas plaform if we need batch requests. There is no sense to use batch with current design.
> My opinion is that we need to remove all batches and use another service that will aggregate requests if we need it.
> Each platform have its own problems. Sometimes it is not working, sometimes it is broken because api changed. We cannot guarantee that with batch request there will not be a common panic


I think this problem can be solved in another PR
And since we have a similar implementation for other services, and it works now, I do not see a problem in this
https://github.com/trustwallet/blockatlas/blob/master/api/registry.go#L80

EnoRage:
> > We need new architecture for all atlas plaform if we need batch requests. There is no sense to use batch with current design.
> > My opinion is that we need to remove all batches and use another service that will aggregate requests if we need it.
> > Each platform have its own problems. Sometimes it is not working, sometimes it is broken because api changed. We cannot guarantee that with batch request there will not be a common panic
> 
> I think this problem can be solved in another PR
> And since we have a similar implementation for other services, and it works now, I do not see a problem in this
> https://github.com/trustwallet/blockatlas/blob/master/api/registry.go#L80

Firstly, we do not need to make anything worse than it is.
Secondly, this issue is need to be solved on backend side

api/endpoint/token.go
prazd: done

vikmeup:
Created a separate issue to improve fetching tokens at any point in the future: https://github.com/trustwallet/blockatlas/issues/1057

PR #1062, opened 2020-04-27
Incremental jsonrpc request id

### LINKED ISSUE #1060: Fix id in jsonrpc.go
It uses `method` as `id`, Looks a bit weird

hewigovens:
Fixes #1060

### DIFF
--- a/pkg/blockatlas/jsonrpc.go
+++ b/pkg/blockatlas/jsonrpc.go
@@ -2,6 +2,8 @@ package blockatlas
 
 import (
 	"encoding/json"
+	"time"
+
 	"github.com/trustwallet/blockatlas/pkg/errors"
 )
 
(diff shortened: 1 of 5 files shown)


vikmeup:
Can we have it incremental?

vikmeup:
Because when you do batch request, you will have same ID, most likely

pkg/blockatlas/jsonrpc.go
```diff
@@ -16,14 +21,14 @@ type (
 		JsonRpc string      `json:"jsonrpc"`
 		Method  string      `json:"method"`
 		Params  interface{} `json:"params,omitempty"`
-		Id      string      `json:"id,omitempty"`
+		Id      int64       `json:"id,omitempty"`
```
catenocrypt, line 24: Overly strict. As RPC may return string ID, I would not limit to int (still generate numeric ID, but as string)

hewigovens: Some upstream rpc doesn’t support string id

PR #1063, opened 2020-04-29
Add Tezos unknown validator

kolya182:
New validators requirements https://github.com/trustwallet/assets/pull/2006  will delist some bakers, in order to unstake XTZ from delisted baker we still need to provide a client with baker address. This PR supplies dummy validator object, similar to https://github.com/trustwallet/blockatlas/commit/af161d518723eec2bcf81efd7b77e343fbff7791

### DIFF
--- a/platform/cosmos/stake.go
+++ b/platform/cosmos/stake.go
@@ -11,7 +11,7 @@ import (
 )
 
 const (
-	lockTime      = 1814400
+	lockTime      = 1814400 // in seconds (21 days)
 	minimumAmount = "1"
 )
 
--- a/platform/tezos/stake.go
+++ b/platform/tezos/stake.go
@@ -31,8 +31,8 @@ func (p *Platform) GetDelegations(address string) (blockatlas.DelegationsPage, e
 func NormalizeDelegation(account Account, validators blockatlas.ValidatorMap) (blockatlas.DelegationsPage, error) {
 	validator, ok := validators[account.Delegate]
 	if !ok {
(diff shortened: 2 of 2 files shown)


platform/tezos/stake.go
```diff
@@ -97,3 +101,22 @@ func normalizeValidator(v Validator) (validator blockatlas.Validator) {
 		Details: getDetails(),
 	}
 }
+
+func getUnknownValidator(address string) blockatlas.StakeValidator {
+	return blockatlas.StakeValidator{
+		ID:     address,
+		Status: false,
+		Info: blockatlas.StakeValidatorInfo{
+			Name:        "Decommissioned",
```
hewigovens, line 110: constant?

platform/cosmos/stake.go
```diff
@@ -11,7 +11,7 @@ import (
 )
 
 const (
-	lockTime      = 1814400
+	lockTime      = 1814400 // in seconds (21 days)
```
catenocrypt, line 14: Having a redundant number in comment is dangerous, can go out of sync.  Instead I suggest:
	lockTime      = 21 * 24 * 3600 // in seconds

platform/tezos/stake.go
kolya182: As a Cosmos, this path should be removed in a future.

platform/cosmos/stake.go
kolya182: I will make this change when will refactor staking platforms. This value doesn't used in any calculation.

Issue #1019, opened 2020-04-08
[Platform-API] Remove CustomAPI for ETH and BTC


EnoRage:
Currently we use Custom API to register routes for ETH and BTC inside platform implementation. 
We need to find a better solution:

A. Try to use TransactionAPI and TokenAPI, TokenTxAPI for ETH and rewrite it to be compatible for this. Also do the same for BTC-like coins. Add New UTXO API specially for BTC-like coins and remove routes from BTC platform implementation

B. Any other option

Why current solution is bad:
1. Platform api must not depends on any http api implementation, it is a separate service that needs to be executed in any form
2. We need to have a consistent code, where it is really easy to understand where there are all the api route paths, we do not need to mix it with other services. Currently there is dependency like that
blockatlas.Platform -> api -> ethereum
But blockatlas.Platform use ethereum package

This is not critical issue, but it is important to use the best practices or at least not using worst of them

catenocrypt:
CustomAPIs removed, routes optimized, PR #1068 .  I plan to continue, to change one route for a service, with coin being a parameter, instead of a route for each coin.

PR #1068, opened 2020-04-30
Refactor: Remove CustomAPI routes for BTC and ETH

catenocrypt:
Issue #1019 .


api/api.go
```diff
@@ -5,16 +5,24 @@ import (
...
 )
 
-func SetupPlatformAPI(root gin.IRouter) {
-	for _, api := range platform.Platforms {
-		RegisterCollectionsAPI(root, api)
-		RegisterTransactionsAPI(root, api)
-		RegisterCustomAPI(root, api)
-		RegisterTokensAPI(root, api)
-		RegisterStakeAPI(root, api)
+func SetupPlatformAPI(router gin.IRouter) {
```
EnoRage, line 8: What is the purpose to split it?

pkg/blockatlas/platform.go
```diff
@@ -23,23 +29,16 @@ type (
...
-		GetTokenListByAddress(address string) (TokenPage, error)
-	}
-
-	// BlockAPI provides block information and lookups
-	BlockAPI interface {
-		Platform
-		CurrentBlockNumber() (int64, error)
-		GetBlockByNumber(num int64) (*Block, error)
+	// TxByAddrAndXPubAPI provides transaction lookup based on address and XPUB (Bitcoin-style)
+	TxByAddrAndXPubAPI interface {
```
EnoRage, line 33: XpubAPI

EnoRage:
Great! Will add couple comments more soon

api/api.go
catenocrypt: A small step in the direction when services are more separated, and handlers are not decided based on interface implementation or not, but explicit lists.  Staking and Naming (as there are only a few of them) have been completely separated.
I plan to change the routes one for each coin, then the for loops here will be gone.

EnoRage: Cannot see any sense. You are just make more code by making 5 loops instead of 1

pkg/blockatlas/errors.go
```diff
@@ -10,3 +10,6 @@ var ErrInvalidAddr = errors.New("invalid address")
 
 // ErrNotFound signals that the resource has not been found
 var ErrNotFound = errors.New("not found")
+
+// ErrInvalidKey signals that the requested key is invalid
+var ErrInvalidKey = errors.New("invalid key")
```
prazd, line 15:
```
var ( 
   ErrNotFound
   ErrInvalidKey
   ...
)
```

platform/ethereum/transaction.go
```diff
@@ -1,45 +1,18 @@
...
+func (p *Platform) GetTokenTxsByAddress(address string, token string) (blockatlas.TxPage, error) {
+	page, err := p.client.GetTransactions(address, p.CoinIndex)
 	if err != nil {
-		logger.Error(err, "Unhandled error")
-		c.AbortWithStatus(http.StatusInternalServerError)
-		return true
+		return nil, err
 	}
-	return false
+	return page, err
```
prazd, line 17: Maybe return `page, nil`?
Or you can delete `if err != nil` check:
```
return p.client.GetTransactions(address, p.CoinIndex)
```

platform/platform.go
```diff
@@ -111,7 +109,22 @@ func getPlatformMap() blockatlas.Platforms {
 	}
 }
 
-func InitCollectionsWhitelist() {
-	CollectionsWhitelist = make(map[uint]bool)
-	CollectionsWhitelist[coin.Ethereum().ID] = true
+func getCollectionsHandlers() map[uint]blockatlas.CollectionsAPI {
```
prazd, line 112: Maybe add alias for this map?

api/registry.go
```diff
@@ -6,143 +6,88 @@ import (
...
-	root.GET("/ns/lookup", endpoint.GetAddressByCoinAndDomain)
-	root.GET("v2/ns/lookup", endpoint.GetAddressByCoinAndDomainBatch)
-}
-
-func RegisterBasicAPI(root gin.IRouter) {
-	root.GET("/", endpoint.GetStatus)
-	root.GET("/metrics", ginprom.PromHandler(promhttp.Handler()))
+func RegisterDomainAPI(router gin.IRouter) {
+	router.GET("/ns/lookup", endpoint.GetAddressByCoinAndDomain)
+	router.GET("v2/ns/lookup", endpoint.GetAddressByCoinAndDomainBatch)
```
prazd, line 87: `/v2/ns/lookup/`

catenocrypt:
I planned to make API routes generic w.r.t. coin, e.g. '/v1/:coin/address/:address' instead of '/v1/' + coin + '/address/:address'.  However, this is not easily possible.  There are many routes that would conflict this way:

- "/v1/:coin/:address"
- "/v1/:coin/address/:address"

and
- "/v2/:coin/transactions/:address"
- "/v2/:coin/staking/validators"
- "/v2/staking/delegations"
- etc.

The only way to make this is through an overly generic API path like "/v1/:arg1/:arg2/:arg3", and dispatch from code.  But this would be not clean, hard to see, hard to document, etc.

 Therefore the correct scheme of per-coin API routes remains
`for (handler := ...)  "/v1/" + handler.coin + "/address/:address/"`

api/api.go
catenocrypt: It's about separation of concerns.  Think like this:

A: Get a bunch of handlers, DECIDE which ones are relevant and REGISTER an path for them
B: Get a list of relevant handlers, REGISTER a path for them.

Separation of concerns sometimes comes with a few extra lines, and I think it worth a few lines.
However, my plan was to change the routes generic w.r.t. coin, in which case the for loops would have disappear completely, but this turned to be not plausible (see prev. comment).  Therefor I may revert this change.

platform/ethereum/transaction.go
catenocrypt: yes, cleaner, thanks

api/registry.go
catenocrypt: I added starting / to 3 routes that missed it, but I'm not sure about the trailing slashes, that's missing in many routes.  What's the significance of trailing slash?

platform/platform.go
catenocrypt: done

catenocrypt:
Still keeping as WIP, I need to test it more thoroughly.

catenocrypt:
API unit tests have been verified together with contents; exact same output is produce before PR.

catenocrypt:
This PR is ready for review now, please review.

EnoRage:
Will take a look today

EnoRage:
PR will be merged with ignored codecov/patch check

PR #1070, opened 2020-05-01
Add /v2/xpub endpoint 

EnoRage:
/v2/xpub endpoint was added by @vikmeup request

Swagger docs were also updated


EnoRage:
@catenocrypt Sorry that I made it at the time when you are refactoring especially that codebase. But we cannot make other teams to wait for the requested changes

catenocrypt:
No problem, noted. 
What is the difference of this new v2 compared to v1, is it the same?

PR #1076, opened 2020-05-07
Sort struct properties in alphabetic order

kolya182:
For the sake of some consistency sort struct fields in alphabetic order

### DIFF
--- a/pkg/blockatlas/tx.go
+++ b/pkg/blockatlas/tx.go
@@ -8,47 +8,47 @@ import (
 
 const (
 	StatusCompleted Status = "completed"
-	StatusPending   Status = "pending"
 	StatusError     Status = "error"
+	StatusPending   Status = "pending"
 
-	DirectionOutgoing Direction = "outgoing"
 	DirectionIncoming Direction = "incoming"
+	DirectionOutgoing Direction = "outgoing"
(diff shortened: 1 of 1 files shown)


pkg/blockatlas/tx.go
```diff
@@ -66,7 +66,7 @@ type (
 	Block struct {
 		Number int64  `json:"number"`
 		ID     string `json:"id,omitempty"`
-		Txs    []Tx   `json:"txs"`
+		Txs    TxPage `json:"txs"`
```
hewigovens, line 69:
```suggestion
		Txs    []Tx `json:"txs"`
```

kolya182: Sorted via https://codebeautify.org/sort-text-lines, but this change let me fix that

EnoRage:
Goland also have this feature

vikmeup:
Can we sort this for the whole project and add this to linting CI? otherwise it's pointless to make such changes as as make git history quite hard to look after

EnoRage:
@vikmeup Good idea, have plans for it

kolya182:
> @vikmeup Good idea, have plans for it

Issue to track https://github.com/trustwallet/blockatlas/issues/1089

PR #1078, opened 2020-05-09
Sort Harmony trx

kolya182:
Critcal Fixes for  https://github.com/trustwallet/rpc-services/issues/309


EnoRage:
We ignore codecov and do not cover more test, it is critical fix

PR #1081, opened 2020-05-10
Sort Txs on api level now

EnoRage:
Cleanup prev PRs. Now we sort all txs on api level, not on provider level. Add tests for it


EnoRage:
I do not change api a lot. We will be rewriting to and split business logic to other service

PR #1087, opened 2020-05-10
[Transactions] Improve fetching logic for transactions

vikmeup:
In current implementation now we have ability to include transactions details to avoid calling /transactions endpoint for each transaction that is being parsed from the block or list of transactions

**What does this break?** 

It breaks fetching memos for `kin` blockchain, which is acceptable at this point to avoid making so many requests. 

Another fix:
- fixing block number for stellar as it was using wrong field in the past. For kin it will stay as 0, instead of wrong number

### DIFF
--- a/platform/stellar/client.go
+++ b/platform/stellar/client.go
@@ -15,6 +15,7 @@ func (c *Client) GetTxsOfAddress(address string) ([]Payment, error) {
 	query := url.Values{
 		"order": {"desc"},
 		"limit": {"25"},
+		"join":  {"transactions"},
 	}
 	path := fmt.Sprintf("accounts/%s/payments", url.PathEscape(address))
 
@@ -24,17 +25,6 @@ func (c *Client) GetTxsOfAddress(address string) ([]Payment, error) {
 		return nil, err
 	}
 	return payments.Embedded.Records, nil
-}
-
-func (c *Client) GetTxHash(id string) (TxHash, error) {
-	path := fmt.Sprintf("transactions/%s", id)
-
-	var hash TxHash
-	err := c.Get(&hash, path, nil)
-	if err != nil {
-		return hash, err
-	}
-	return hash, nil
 }
 
 func (c *Client) CurrentBlockNumber() (int64, error) {
@@ -63,6 +53,7 @@ func (c *Client) GetBlockByNumber(num int64) (*Block, error) {
 	query := url.Values{
 		"order": {"desc"},
 		"limit": {"100"},
+		"join":  {"transactions"},
 	}
 	path := fmt.Sprintf("ledgers/%d/payments", num)
 
--- a/platform/stellar/transaction.go
+++ b/platform/stellar/transaction.go
@@ -16,13 +16,13 @@ func (p *Platform) GetTxsByAddress(address string) (blockatlas.TxPage, error) {
 	return p.NormalizePayments(payments), nil
 }
 
-func (p *Platform) NormalizePayments(payments []Payment) (txs []blockatlas.Tx) {
+func (p *Platform) NormalizePayments(payments []Payment) []blockatlas.Tx {
+	txs := make([]blockatlas.Tx, 0, len(payments))
 	for _, payment := range payments {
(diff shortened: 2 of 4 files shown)


platform/stellar/transaction.go
```diff
@@ -19,42 +17,17 @@ func (p *Platform) GetTxsByAddress(address string) (blockatlas.TxPage, error) {
 }
 
 func (p *Platform) NormalizePayments(payments []Payment) (txs []blockatlas.Tx) {
```
EnoRage, line 19: func (p *Platform) NormalizePayments(payments []Payment) (txs []blockatlas.Tx) {

TO

func (p *Platform) NormalizePayments(payments []Payment) []blockatlas.Tx {
txs := make([]blockatlas.Tx, 0, len(payments))

?

platform/stellar/transaction.go
```diff
@@ -19,42 +17,17 @@ func (p *Platform) GetTxsByAddress(address string) (blockatlas.TxPage, error) {
...
+		}
 	}
 
-	return
+	return txs
 }
 
 // Normalize converts a Stellar-based transaction into the generic model
-func Normalize(payment *Payment, nativeCoinIndex uint, hash TxHash) (tx blockatlas.Tx, ok bool) {
+func Normalize(payment *Payment, nativeCoinIndex uint) (tx blockatlas.Tx, ok bool) {
```
EnoRage, line 30: Same as at prev comment
(tx blockatlas.Tx, ok bool) => (blockatlas.Tx, bool) ?

platform/stellar/transaction.go
vikmeup: Done

platform/stellar/client.go
```diff
@@ -26,17 +27,6 @@ func (c *Client) GetTxsOfAddress(address string) ([]Payment, error) {
...
-
-	var hash TxHash
-	err := c.Get(&hash, path, nil)
-	if err != nil {
-		return hash, err
-	}
-	return hash, nil
-}
-
 func (c *Client) CurrentBlockNumber() (int64, error) {
```
EnoRage, line 30: make it currentBlockNumber(), because it is not used out of stellar package?

platform/stellar/transaction.go
vikmeup: I don't think this works, as developer won't know what is `bool`, in that case it's preferred to define as `ok bool`

platform/stellar/client.go
```diff
@@ -15,6 +15,7 @@ func (c *Client) GetTxsOfAddress(address string) ([]Payment, error) {
 	query := url.Values{
```
EnoRage, line 15: make it getTxsOfAddress(), because it is not used out of stellar package?

platform/stellar/client.go
vikmeup: I think the goal of the client to expose methods so anyone else can use it

platform/stellar/transaction.go
EnoRage: it is just styling of functions, there is no difference for the developer

EnoRage: Also Normalize => normalize

platform/stellar/transaction.go
```diff
@@ -18,43 +16,18 @@ func (p *Platform) GetTxsByAddress(address string) (blockatlas.TxPage, error) {
 	return p.NormalizePayments(payments), nil
 }
 
-func (p *Platform) NormalizePayments(payments []Payment) (txs []blockatlas.Tx) {
-	var (
-		wg      sync.WaitGroup
-		txsChan = make(chan blockatlas.Tx, len(payments))
-	)
-
+func (p *Platform) NormalizePayments(payments []Payment) []blockatlas.Tx {
```
EnoRage, line 19: NormalizePayments => normalizePayments

vikmeup:
@EnoRage created an issue to track this: https://github.com/trustwallet/blockatlas/issues/1088 to address in different PR for all platforms

Issue #1082, opened 2020-05-10
Remove strange ports implementation from Makefile


EnoRage:
It is really weird implementation of mocked tests run at Makefile

I suppose not to check port by passing some strange empty json at postman/ directory

EnoRage:
@catenocrypt can you take it on yourself, please

catenocrypt:
Please explain in more detail what is wrong and/or propose alternative solution.

EnoRage:
https://github.com/trustwallet/blockatlas/blob/master/tests/postman/mock-healthcheck_data.json

EnoRage:
@catenocrypt what is this?

EnoRage:
https://github.com/trustwallet/blockatlas/blob/master/Makefile#L138
Same question

catenocrypt:
Closed via #1090 .

PR #1090, opened 2020-05-11
Remove empty test iteration file for mock startup.

catenocrypt:
Reduce strangeness factor, addresses #1082 . Test only.


catenocrypt:
Codecov diff is mixed up, shows results for **another** pull request...

image

PR #1091, opened 2020-05-11
Enable ethereum pending txs

hewigovens:
Changes:
1. Remove pending tx condition
2. Check data field when fill metadata
3. Calculate gas (limit * price) for pending tx
4. Add pending incoming eth/token transfer test

### DIFF
--- a/platform/ethereum/blockbook/model.go
+++ b/platform/ethereum/blockbook/model.go
@@ -73,4 +73,5 @@ type EthereumSpecific struct {
 	GasLimit *big.Int `json:"gasLimit"`
 	GasUsed  *big.Int `json:"gasUsed"`
 	GasPrice string   `json:"gasPrice"`
+	Data     string   `json:"data,omitempty"`
 }
--- a/platform/ethereum/blockbook/model_extension.go
+++ b/platform/ethereum/blockbook/model_extension.go
@@ -1,6 +1,8 @@
 package blockbook
 
 import (
+	"math/big"
+
 	"github.com/trustwallet/blockatlas/pkg/blockatlas"
 )
 
@@ -22,6 +24,21 @@ func (t *Transaction) FromAddress() string {
 		return t.Vin[0].Addresses[0]
 	}
 	return ""
+}
+
+func (t *Transaction) GetFee() string {
+	status, _ := t.EthereumSpecific.GetStatus()
+	if status != blockatlas.StatusPending {
+		return t.Fees
+	}
+
+	gasLimit := t.EthereumSpecific.GasLimit
+	gasPrice, ok := new(big.Int).SetString(t.EthereumSpecific.GasPrice, 10)
+	if gasLimit == nil || !ok {
+		return "0"
+	}
+	fee := new(big.Int).Mul(gasLimit, gasPrice)
+	return fee.String()
 }
 
 func (t *Transaction) ToAddress() string {
--- a/platform/ethereum/blockbook/transaction_test.go
+++ b/platform/ethereum/blockbook/transaction_test.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/trustwallet/blockatlas/pkg/blockatlas"
 )
 
@@ -139,7 +140,7 @@ func TestNormalizePage(t *testing.T) {
 						}
 					},
 					{
-						"txid": "0xfea3444def47c77c5ede0f705d1bc84bb258d51fa454971985c181d6bc679144",
+						"txid": "0x7a3929f2fad5e61f535ed5c1317f34e739655d582bc1b0161b9869b0957df6af",
 						"vin": [
 							{
 								"n": 0,
@@ -151,7 +152,7 @@ func TestNormalizePage(t *testing.T) {
 						],
 						"vout": [
 							{
-								"value": "234000000000000",
+								"value": "567000000000000",
 								"n": 0,
 								"addresses": [
 									"0x47331175b23C2f067204B506CA1501c26731C990"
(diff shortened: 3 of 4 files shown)


platform/ethereum/blockbook/transaction_test.go
```diff
@@ -182,6 +217,23 @@ func TestNormalizePage(t *testing.T) {
...
+					"id": "0xfea3444def47c77c5ede0f705d1bc84bb258d51fa454971985c181d6bc679144",
+					"coin": 60,
+					"from": "0x7d8bf18C7cE84b3E175b339c4Ca93aEd1dD166F1",
+					"to": "0x47331175b23C2f067204B506CA1501c26731C990",
+					"fee": "0",
+					"date": 1589249442,
+					"block": 0,
+					"status": "pending",
+					"sequence": 0,
+					"type": "contract_call",
```
vikmeup, line 230: why does it says contract call?

platform/ethereum/blockbook/transaction_test.go
```diff
@@ -182,6 +217,23 @@ func TestNormalizePage(t *testing.T) {
 						"from": "0x7d8bf18C7cE84b3E175b339c4Ca93aEd1dD166F1",
 						"to": "0x7d8bf18C7cE84b3E175b339c4Ca93aEd1dD166F1"
 					}
+				  },{
+					"id": "0xfea3444def47c77c5ede0f705d1bc84bb258d51fa454971985c181d6bc679144",
+					"coin": 60,
+					"from": "0x7d8bf18C7cE84b3E175b339c4Ca93aEd1dD166F1",
+					"to": "0x47331175b23C2f067204B506CA1501c26731C990",
+					"fee": "0",
```
vikmeup, line 225: can we include fee here? as gas price * gas limit

hewigovens:
@vikmeup current blockbook api returns pending tx like this
```json
{
    "txid": "0xf1fc796cab0a025775658c41b7a77556528081a4141a3dcc821e4c60d8fcf9cc",
    "vin": [
        {
            "n": 0,
            "addresses": [
                "0x23a80893B3CE3B17598c2077739559B5f13F81bA"
            ],
            "isAddress": true
        }
    ],
    "vout": [
        {
            "value": "0",
            "n": 0,
            "addresses": [
                "0xdAC17F958D2ee523a2206206994597C13D831ec7"
            ],
            "isAddress": true
        }
    ],
    "blockHeight": -1,
    "confirmations": 0,
    "blockTime": 1589251274,
    "value": "0",
    "fees": "0",
    "rbf": true,
    "tokenTransfers": [
        {
            "type": "ERC20",
            "from": "0x23a80893B3CE3B17598c2077739559B5f13F81bA",
            "to": "0x87c574FB2DF0EaA2dAf5fc4a8A16dd3Ce39011B1",
            "token": "0xdAC17F958D2ee523a2206206994597C13D831ec7",
            "name": "Tether USD",
            "symbol": "USDT",
            "decimals": 6,
            "value": "3940610000"
        }
    ],
    "ethereumSpecific": {
        "status": -1,
        "nonce": 0,
        "gasLimit": 100000,
        "gasUsed": null,
        "gasPrice": "30000000000"
    }
}
```

hewigovens:
we couldn't get nonce/data/gas, so the type is fallback to contract_call

vikmeup:
Can we adjust logic to support this? since we have all the data?

platform/ethereum/blockbook/transaction_test.go
hewigovens: this one can be fixed

hewigovens:
need to include abi parsing somehow, not a quick fix, also we need to update to latest code to include data field

platform/ethereum/blockbook/transaction_test.go
hewigovens: done

platform/ethereum/blockbook/transaction_test.go
hewigovens: this test case is correct now, data: `0xdeadbeef`

hewigovens:
> Can we adjust logic to support this? since we have all the data?

it turns out we don't need to do abi parsing now, the only issue left is low gas price tx might be dropped (but can be seen in etherscan.io)

platform/ethereum/blockbook/model.go
```diff
@@ -73,4 +73,5 @@ type EthereumSpecific struct {
 	GasLimit *big.Int `json:"gasLimit"`
 	GasUsed  *big.Int `json:"gasUsed"`
 	GasPrice string   `json:"gasPrice"`
+	Data     string   `json:"data,omitempty"`
```
EnoRage, line 76: It could be empty?

platform/ethereum/blockbook/model_extension.go
```diff
@@ -24,6 +26,21 @@ func (t *Transaction) FromAddress() string {
 	return ""
 }
 
+func (t *Transaction) GetFee() string {
+	status, _ := t.EthereumSpecific.GetStatus()
```
EnoRage, line 30: can you handle error here please

platform/ethereum/blockbook/model.go
hewigovens: Yes, this field is added recently, we may switch to old node which doesn’t return data

platform/ethereum/blockbook/model_extension.go
hewigovens: Looks weird, but _ is not an error but status string, can be improved tho

platform/ethereum/blockbook/model_extension.go
```diff
@@ -24,6 +26,21 @@ func (t *Transaction) FromAddress() string {
...
+	if status != blockatlas.StatusPending {
+		return t.Fees
+	}
```
vikmeup, line 33: can you add test for this specific function?

hewigovens: I think it's already covered in `TestNormalizePage`, we have both pending and completed txs

PR #1097, opened 2020-05-13
Checksum address according to Wanchain EIP55 spec

kolya182:
PR aims to solve issue when Wanchain transfers direction incorrectly displayed in Atlas response

### DIFF
--- a/pkg/address/address.go
+++ b/pkg/address/address.go
@@ -59,3 +59,30 @@ func HexToAddress(hexAddr string) (b58 string, err error) {
 	b58 = base58.EncodeAlphabet(bytes, base58.BTCAlphabet)
 	return
 }
+
+// Returns an EIP55 Wanchain compliant hex string representation of the address.
+// See https://wandevs.org/docs/difference-between-wanchain-and-ethereum/
+// https://github.com/wanchain/go-wanchain/blob/b238c203ca7dc6a578d57c0c473bec0fcf2431bd/common/types.go#L173
+func EIP55ChecksumWanchain(address string) string {
+	v := []byte(Remove0x(strings.ToLower(address)))
+	sha := sha3.NewLegacyKeccak256()
+	_, err := sha.Write(v)
+	if err != nil {
+		logger.Error(err)
+	}
+	hash := sha.Sum(nil)
+
+	result := v
+	for i := 0; i < len(result); i++ {
+		hashByte := hash[i/2]
+		if i%2 == 0 {
+			hashByte = hashByte >> 4
+		} else {
+			hashByte &= 0xf
+		}
+		if result[i] > '9' && hashByte <= 7 {
+			result[i] -= 32
+		}
+	}
+	return "0x" + string(result)
+}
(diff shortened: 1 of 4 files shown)


pkg/address/address.go
```diff
@@ -59,3 +59,30 @@ func HexToAddress(hexAddr string) (b58 string, err error) {
...
+// Returns an EIP55 Wanchain compliant hex string representation of the address.
+// See https://wandevs.org/docs/difference-between-wanchain-and-ethereum/
+// https://github.com/wanchain/go-wanchain/blob/b238c203ca7dc6a578d57c0c473bec0fcf2431bd/common/types.go#L173
+func EIP55ChecksumWanchain(address string) string {
+	v := []byte(Remove0x(strings.ToLower(address)))
+	sha := sha3.NewLegacyKeccak256()
+	_, err := sha.Write(v)
+	if err != nil {
+		logger.Error(err)
+	}
```
catenocrypt, line 72: Early return here?

EnoRage:
What about not wanchain responses? What was the issue with it?

pkg/address/address.go
kolya182: `""` or `err` ?

kolya182:
> What about not wanchain responses? What was the issue with it?

Amount all ETH forks only Wanchain generates checksum differently, see comment above `EIP55ChecksumWanchain`

EnoRage:
Can we verify that other forks will not have any problems with this normalization?

pkg/address/address.go
kolya182: In case if we handle we need also add err handling here https://github.com/trustwallet/blockatlas/blob/master/pkg/address/address.go#L27

hewigovens: I think we can just return original string if there is an error, same for `EIP55Checksum`

platform/ethereum/trustray/transaction.go
```diff
@@ -59,26 +60,35 @@ func AppendTxs(in []blockatlas.Tx, srcTx *Doc, coinIndex uint) (out []blockatlas
...
 	}
 	op := &srcTx.Ops[0]
-
+	// Token transfer transaction
 	if op.Type == blockatlas.TxTokenTransfer && op.Contract != nil {
-		tokenTx := baseTx
+		var tokenID string
+		switch coinIndex {
+		case coin.WAN:
+			tokenID = address.EIP55ChecksumWanchain(op.Contract.Address)
```
hewigovens, line 75: Add a helper `func normalizeAddress(string, coin)`? so we don't need to have switch cases in L73 and L122

kolya182: https://github.com/trustwallet/blockatlas/pull/1097/commits/08118e4648ee32e98cb6ee1bc71885b6be1445d9

kolya182:
> Can we verify that other forks will not have any problems with this normalization?

added test

pkg/address/address.go
kolya182: Left as in `EIP55Checksum`

catenocrypt: I would definitely return original string after logging in this error branch.  If sha.Write() fails, who knows what will sha.Sum() return?  Maybe empty array --> then we will overrun, or maybe all zeroes --> we will generate wrong result.

PR #1098, opened 2020-05-14
[Waves] Remove redundant code, fix styling

kolya182:


platform/waves/client.go
```diff
@@ -2,35 +2,37 @@ package waves
...
+	if err != nil {
+		return nil, err
+	}
 
 	if len(txs) > 0 {
-		return txs[0], err
+		return txs[0], nil
 	} else {
-		return []Transaction{}, err
+		return []Transaction{}, nil
```
prazd, line 24: return nil, nil

PR #923, opened 2020-02-24
[Harmony] Adding get validators and get delegations #733

coolcottontail:
Moved gupadhyaya 's old PR (https://github.com/trustwallet/blockatlas/pull/733) here

PR #733, opened 2020-01-21
[Harmony] Adding get validators and get delegations

gupadhyaya:

PR #1099, opened 2020-05-15
Add harmony staking support in mainnet

coolcottontail:
This is resubmit for old PRs (#733 and #923) which supports harmony staking in block atlas.
The old PR failed test because the Harmony Mainnet didn't have the staking RPC (and I can't use testnet endpoints).

Now I can re-submit this PR since Harmony MainNet has staking RPC ready. 

Test status:
1) passed local test
2) passed the tests which failed in the previous PRs:
1. cd cmd/platform_api
2. ATLAS_PLATFORM=harmony go run main.go
3. Try these requests:

http://localhost:8420/v2/harmony/staking/validators
http://localhost:8420/v2/harmony/staking/delegations/one1e4mr7tp0a76wnhv9xd0wzentdjnjnsh3fwzgfv

### DIFF
--- a/platform/harmony/stake.go
+++ b/platform/harmony/stake.go
@@ -1,0 +1,123 @@
+package harmony
+
+import (
+	"github.com/trustwallet/blockatlas/pkg/blockatlas"
+	"github.com/trustwallet/blockatlas/pkg/errors"
+	"github.com/trustwallet/blockatlas/pkg/logger"
+	services "github.com/trustwallet/blockatlas/services/assets"
+	"math/big"
+	"strconv"
+)
+
+const (
+	lockTime  = 604800 // in seconds (7 epochs or 7 days)
+)
+
+func (p *Platform) GetValidators() (blockatlas.ValidatorPage, error) {
+	results := make(blockatlas.ValidatorPage, 0)
+	validators, err := p.client.GetValidators()
+	if err != nil {
+		return results, err
+	}
+
+	for _, v := range validators.Validators {
+		var apr float64
+		if apr, err = strconv.ParseFloat(v.Lifetime.Apr, 64); err != nil {
+			apr = 0
+		}
+		results = append(results, normalizeValidator(v, apr))
+	}
+
+	return results, nil
+}
+
+func (p *Platform) GetDetails() blockatlas.StakingDetails {
+	apr := p.GetMaxAPR()
+	return getDetails(apr)
+}
+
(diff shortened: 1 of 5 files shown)


platform/harmony/transaction.go
```diff
@@ -8,6 +8,8 @@ import (
 	"strconv"
 )
 
+const Annual = 10
```
kolya182, line 11: Is any better way to calculate APR ?

platform/harmony/stake.go
```diff
@@ -0,0 +1,123 @@
...
+		}
+		results = append(results, delegation)
+	}
+	return results
+}
+
+func getDetails(apr float64) blockatlas.StakingDetails {
+	return blockatlas.StakingDetails{
+		Reward:        blockatlas.StakingReward{Annual: apr},
+		MinimumAmount: blockatlas.Amount("0"),
```
kolya182, line 111: What does protocol say about minimum amount ?.

coolcottontail: this is 1000 ONE. I fixed it.

platform/harmony/transaction.go
coolcottontail: This is only used as fallback Apr if there's no validator available.

The APR is calculated  line 43 of platform/harmony/stake.go,  we go through all returned active validators and get the validator with maximal APR and use it as the current APR. 

https://github.com/trustwallet/blockatlas/pull/1099/files#diff-c861e736adc338bed48219a604ff84b7R43

func (p *Platform) GetMaxAPR() float64 {
	validators, err := p.client.GetValidators()
	if err != nil {
		logger.Error("GetMaxAPR", logger.Params{"details": err, "platform": p.Coin().Symbol})
		return Annual
	}

	var max = 0.0
	for _, e := range validators.Validators {
		var apr float64
		if apr, err = strconv.ParseFloat(e.Lifetime.Apr, 64); err != nil {
			apr = 0.0
		}

		if apr > max {
			max = apr
		}
	}

	return max
}

catenocrypt:
Overall coverage is decreasing, diff coverage is low.  Could you increase the coverage of the new code?

coolcottontail:
Is there a way to add more coverage to the code ?  I tested most of the code except those require RPCCall to network, which requires some mock . I am trying to find an example but I found there's no other platform coins doing RPC mock...

coolcottontail:
can someone help to approve and merge ?

Issue #1106, opened 2020-05-18
KIN empty amount during blocks parsing


EnoRage:
Sometimes getBlockByNumber parse txs with empty amount it breaks to UnmarshalJSON for blockatlas.Tx

catenocrypt:
In platform/stellar/transaction.go there is explicit check for Amount != "", though error from DecimalToSatoshis is swallowed.  However, there is a bug in DecimalToSatoshis: leftmost 0s are trimmed, but if input is "0", result is "" (and no error).  In this case in normalized tx Value will be "".

EnoRage:
👍🏻 great, just need to fix it

catenocrypt:
DecimalToSatoshis is used by Cosmos and Solana.

catenocrypt:
Should be solved, #1110 .

PR #1110, opened 2020-05-19
Fix DecimalToSatoshis for 0 input.

catenocrypt:
DecimalToSatoshis has a bug: for input "0", it trims the single 0 to "", and returns as a valid response.  Results in unparseable normalized TX in case of Solana, KIN, Cosmos.  See #1106 . 

Also solve a minor sporadic unit test, #1111 .

### DIFF
--- a/platform/cosmos/transaction.go
+++ b/platform/cosmos/transaction.go
@@ -146,9 +146,13 @@ func (p *Platform) fillTransfer(tx *blockatlas.Tx, transfer MessageValueTransfer
 }
 
 func (p *Platform) fillDelegate(tx *blockatlas.Tx, delegate MessageValueDelegate, events Events, msgType TxType) {
-	value, err := numbers.DecimalToSatoshis(delegate.Amount.Quantity)
-	if err != nil {
-		return
+	value := ""
+	if len(delegate.Amount.Quantity) > 0 {
+		var err error
+		value, err = numbers.DecimalToSatoshis(delegate.Amount.Quantity)
+		if err != nil {
+			return
+		}
 	}
 	tx.From = delegate.DelegatorAddr
 	tx.To = delegate.ValidatorAddr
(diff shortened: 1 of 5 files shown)


platform/nimiq/transaction_test.go
```diff
@@ -68,7 +69,7 @@ var (
 		From:  "NQ74 SJ0Q 49T1 4XQL KABH 1RUC 8DPT 9F0U 9P0B",
 		To:    "NQ97 18BJ 33YV QGHQ BV2K 56V4 12CH J8TD S9S3",
 		Fee:   "300",
-		Date:  time.Now().Unix(),
+		Date:  666666, // special placholder value
```
hewigovens, line 72: why not just hardcode a timestamp?

platform/cosmos/transaction.go
```diff
@@ -146,9 +146,13 @@ func (p *Platform) fillTransfer(tx *blockatlas.Tx, transfer MessageValueTransfer
...
-	value, err := numbers.DecimalToSatoshis(delegate.Amount.Quantity)
-	if err != nil {
-		return
+	value := ""
+	if len(delegate.Amount.Quantity) > 0 {
+		var err error
+		value, err = numbers.DecimalToSatoshis(delegate.Amount.Quantity)
+		if err != nil {
+			return
+		}
```
hewigovens, line 155: maybe we can move this to another func (2 if statements)?  normalizeValue/Amount

platform/nimiq/transaction_test.go
catenocrypt: Product code fills missing with current time, I did not want to change that.

platform/cosmos/transaction.go
catenocrypt: it is context-specific where value can be empty/missing, and where not.

Issue #1107, opened 2020-05-18
PG - Limit amount of bulk create objects


EnoRage:
Recommend to change bulk create to package with bulk create

PR #1113, opened 2020-05-20
[BulkInsert] Use gorm-bulk-insert instead of BulkCreate

prazd:


db/models/subscriptions.go
```diff
@@ -11,6 +11,6 @@ type Subscription struct {
 type SubscriptionData struct {
 	ID             uint   `gorm:"primary_key;"`
 	SubscriptionId uint   `gorm:"primary_key; column:subscription_id; auto_increment:false"`
-	Coin           uint   `gorm:"primary_key; column:coin; auto_increment:false"`
+	Coin           *uint  `gorm:"primary_key; column:coin; auto_increment:false"`
```
EnoRage, line 14: Will be changed as well as SubscriptionData table a bit later.

EnoRage:
https://github.com/trustwallet/blockatlas/issues/1107

Issue #1065, opened 2020-04-29
[Observer] Remove GUID field 


EnoRage:
I suggest to remove guid field for blockatlas subscriptions. There is no need at it. 
We suggest that client which will subscribe - will save information about person with that address. So, there is no need to store a specific guid for addresses. We can just use addresses instead.

EnoRage:
@vikmeup @prazd what do you think about that?

vikmeup:
guid help us identify association between entity <> address, maybe ok to keep as is for now?

PR #1115, opened 2020-05-20
Change subscriptions table and the way of storing it

EnoRage:


EnoRage:
https://github.com/trustwallet/blockatlas/issues/1065

catenocrypt:
I don't fully understand the things behind it, but I approve.
Diff codecov 0% : if measurement is real, doesn't look good.

Issue #1112, opened 2020-05-20
[Platform] Add Lending API, use Staking with Lending as Finance api


EnoRage:
We need to add Lending api (same as staking)

We currently needs to support Compound for lending.

Also there is some research about other lending providers that can be integrated.

EnoRage:
It is still issue for open discussion, we need specification for clients and action flow for lending

EnoRage:
@catenocrypt can you take it after  current issue? What do you think?

EnoRage:
I suggest:

https://aave.com/

https://compound.finance/

https://iearn.finance/earn

https://idle.finance/#/dashboard/best

catenocrypt:
Lending API prototype is being worked on here: #1116 .

vikmeup:
Not relevant anymore

PR #1116, opened 2020-05-25
WIP Lending API prototype

catenocrypt:
Standalone prototype for Lending API, WIP, #1112 .


EnoRage:
Cannot understand why do you have mocks? I think, that the most important part of compound research is using real api / contracts to fetch data. Isn’t it?

EnoRage:
https://github.com/bonedaddy/go-compound

EnoRage:
We need:
1. Get information about each Ctoken data (APR, min amount, etc). to supply
2. Get information about already supplied ctoken for specific address

That’s it

EnoRage:
Screenshot 2020-05-30 at 03 40 15
We need to fetch that data

catenocrypt:
The mock is just a temporary, a scaffolding to get a better grip on the API.

The structure is:
- Client to connect to Compound API
- Compound Provider having an internal API, using the client
- API exposing lending API, routing to provider (only one for now), doing some normalization
Very similar to other providers in blockatlas.

PR #1117, opened 2020-05-27
Change gorm option for subscriptions createdAt field

EnoRage:
This optimization reduce INSERT time 1.5x for 3000 objects locally. Hop it will be close to the local on prod


EnoRage:
Codecov patch ignored

PR #1128, opened 2020-06-03
Update swagger files and readme

hewigovens:
Changes:

- Add `go-swagger` usage, view api docs without running blockatlas
- Run `make go-gen-docs`


EnoRage:
New docs doesn’t have v2/ endpoint and xpub endpoint

hewigovens:
that's weird, annotations are gone?

EnoRage:
> that's weird, annotations are gone?

Yep, it is the legacy of old api architecture

hewigovens:
removed legacy v1 docs

EnoRage:
I think we still need old swagger docs, let's just keep the info about new command

EnoRage:
We need to have /v1, /v2 docs

hewigovens:
@EnoRage v1 docs are back, just update some ids

PR #1139, opened 2020-06-09
Mockserver: -update option to update data files from external services.

catenocrypt:
mockserver -update   updates all external data files with ExtURL set, from external services

### DIFF
--- a/mock/mockserver/mockserver.go
+++ b/mock/mockserver/mockserver.go
@@ -3,21 +3,24 @@
 package main
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
-	"gopkg.in/yaml.v2"
 	"io/ioutil"
 	"log"
 	"net/http"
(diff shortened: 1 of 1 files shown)


EnoRage:
What is the reason to do it?

catenocrypt:
Sometimes it happens that external service changes, or we switch to new version of it, like when @kolya182 had to manually change the data files due to binance API.  This is a utility to do it automatically; of course updated files have to be tested and reviewed.

EnoRage:
> Sometimes it happens that external service changes, or we switch to new version of it, like when @kolya182 had to manually change the data files due to binance API. This is a utility to do it automatically; of course updated files have to be tested and reviewed.

You just need to paste new json into file for that. If it doesn’t works like that - it must. Because it was discussed previously

catenocrypt:
Closing, baffled.

PR #1140, opened 2020-06-09
API errors handling changes

EnoRage:
- Remove models package
- Use AbortWithJSON
- Small cleanup as well


catenocrypt:
config.yml update needed to configmock.yml too (I think that causes failed test)

EnoRage:
@maxUo

Issue #1040, opened 2020-04-20
Make list of bakers dynamic


kolya182:
## Current Functionality
Blockatlas gets list of bakers from Assets repo, at some point in time baker can run out of capacity to accept staking and users would continue delegate with out getting reward out. Some bakers come and [change](https://github.com/trustwallet/assets/pull/1930) their status, some don't. This is not sustainable and we need to find the way check capacity of each baker dynamically

## Possible Improvement
https://mainnet-tezos.giganode.io/chains/main/blocks/head/context/delegates/tz1V3yg82mcrPJbegqVCPn6bC8w1CSTRp3f8

vikmeup:
This information is available from the rpc endpoint: `/chains/main/blocks/head/context/delegates/tz1V3yg82mcrPJbegqVCPn6bC8w1CSTRp3f8` 

From the team:
```
...GET ../<block_id>/context/delegates/<pkh>
Returns staking balances and whether deactivated or not
```

EnoRage:
What is the status on it? @kolya182

vikmeup:
@EnoRage This is completed?

EnoRage:
Nope

EnoRage:
#1153

PR #1153, opened 2020-06-19
Change stake api (Tezos)

EnoRage:
#1040


vikmeup:
@EnoRage if this deployed, we should remove this from assets: https://github.com/trustwallet/assets/blob/master/blockchains/tezos/validators/list.json#L405

and make sure script is removed from assets that was adding that data

PR #1155, opened 2020-06-23
Naming: rename all to naming (not domain); renames only.

catenocrypt:


EnoRage:
What is the purpose to rename crypto domain assets service (FIO **domain**, Unstoppable **domains** and etc) to naming? I think it is more logical to have domain. We have TLD function, we have subdomains, we have domain zone like .eth

catenocrypt:
> What is the purpose to rename crypto domain assets service (FIO **domain**, Unstoppable **domains** and etc) to naming? I think it is more logical to have domain. We have TLD function, we have subdomains, we have domain zone like .eth

Consistency.  In some places we call it Naming, like NamingAPI.  The biggest one is ENS, Ethereum Naming Service; FIO does not call it domain, but FIO Addresses, in the API paths we use '/ns/lookup' where ns is naming service, naming is more intuitive for this service, that's all.

EnoRage:
We are using service to work specially with domains. We resolve ENS **domain**. Naming is a bad and to much abstract naming for such case

catenocrypt:
Closing

PR #1184, opened 2020-08-07
[API/Binance] Change client implementation, remove explorer api

EnoRage:


platform/binance/token.go
```diff
@@ -1,65 +1,17 @@
...
-	"github.com/trustwallet/blockatlas/coin"
 	"github.com/trustwallet/blockatlas/pkg/blockatlas"
-	"strings"
 )
 
 func (p *Platform) GetTokenListByAddress(address string) (blockatlas.TokenPage, error) {
-	account, err := p.rpcClient.fetchAccountMetadata(address)
+	account, err := p.client.FetchAccountMeta(address)
 	if err != nil || len(account.Balances) == 0 {
 		return []blockatlas.Token{}, nil
```
prazd, line 10: return nil instead of []blockatlas.Token{}

platform/binance/transaction.go
```diff
@@ -3,198 +3,20 @@ package binance
...
-	"strconv"
-	"strings"
-	"sync"
 )
 
-const emptyToken = ""
-
 func (p *Platform) GetTxsByAddress(address string) (blockatlas.TxPage, error) {
-	explorerResponse, err := p.GetTokenTxsByAddress(address, emptyToken)
+	txsFromClient, err := p.client.FetchTransactionsByAddressAndAssetID(address, coin.Binance().Symbol)
```
prazd, line 9: FetchTransactionsByAddressAndAssetID - but params `address` and `coin.Binance().Symbol`
Symbol != assetID?

EnoRage: Token ID will be better 👍

PR #1218, opened 2020-09-03
[DB] Replaces subquery by Joins in GetSubscriptionsForNotifications.

jeyldii:


jeyldii:
Make tests with gorm v2. If we update gorm we can optimize x2 every queries like in GetSubscriptionsForNotifications

db/notification.go
```diff
@@ -49,44 +36,38 @@ func (i *Instance) AddSubscriptionsForNotifications(addresses []string, ctx cont
...
-		Select("id").
-		Where("address in (?)", addresses).
-		Limit(len(addresses)).
-		QueryExpr()
-
-	return db.Where("address_id in (?)", addressSubQuery).
-		Delete(&models.NotificationSubscription{}).
-		Limit(len(addresses)).
-		Error
+	q := `DELETE FROM notification_subscriptions ns USING addresses a where ns.address_id = a.id AND a.address IN (?);`
```
vikmeup, line 71: is there a way to use gorm delete function to avoid plain sql queries?

db/models/address.go
```diff
@@ -2,7 +2,7 @@ package models
 
 type Address struct {
 	ID      uint   `gorm:"primary_key"`
-	Address string `gorm:"type:varchar(128); unique_index"`
+	Address string `gorm:"type:varchar(128); index:unique"`
```
vikmeup, line 5: `index:unique` does this work or did previous version worked? I think it might need to be `sql:"index"` ?

vikmeup: In other examples it was: https://github.com/trustwallet/blockatlas/pull/1218/files#diff-07e9c6898f34339b8de0bd692b136955R7

db/db.go
```diff
@@ -1,115 +1,114 @@
...
+			log.New(os.Stdout, "\r\n", log.LstdFlags),
+			gormlogger.Config{
+				SlowThreshold: time.Second,
+				LogLevel:      gormlogger.Info,
+				Colorful:      false,
+			},
+		)
 	}
-
+	g, err := gorm.Open(postgres.Open(uri), cfg)
```
vikmeup, line 35:
```suggestion
	db, err := gorm.Open(postgres.Open(uri), cfg)
```

for consistency let's use `db` as in other places

db/tracker_test.go
```diff
@@ -3,36 +3,48 @@ package db
...
 	assert.Nil(t, i.SetLastParsedBlockNumber("bitcoin", 1, context.Background()))
+	if err = baseDB.Close(); err != nil {
+		log.Fatal(err)
+	}
 }
 
 func TestHeightBlockMap_GetHeight(t *testing.T) {
 	db, mock := setupDB(t)
-	defer db.Close()
+	baseDB, err := db.DB()
```
vikmeup, line 37: can we add this code into `setupDB` ? to avoid duplicating for each test case

db/models/address.go
EnoRage: > In other examples it was: https://github.com/trustwallet/blockatlas/pull/1218/files#diff-07e9c6898f34339b8de0bd692b136955R7

It is correct. We need to have an index. Also it is unique index. Gorm v2 has new type declarations

vikmeup: Do you have a link to new type of declarations?

EnoRage:
@jeyldii 
1. please remove all todos here as comments in github repo.
2. for each change not connected with gorm - add new pr

db/models/address.go
jeyldii: I changed it to ```uniqueIndex``` which equal index, unique. 

Declaring Models doc:  https://gorm.io/docs/models.html

cmd/api/main.go
```diff
@@ -56,9 +61,6 @@ func init() {
 			logger.Fatal(err)
 		}
 		t = tokensearcher.Init(database, platform.TokensAPIs, mq.TokensRegistration)
-
-		go db.RestoreConnectionWorker(database, time.Second*10, pgUri)
```
EnoRage, line 60: move fatal worker after init of mq and pg

cmd/parser/main.go
```diff
@@ -65,14 +62,17 @@ func init() {
...
+	logMode := viper.GetBool("postgres.log")
 	var err error
-	database, err = db.New(pgUri, pgReadUri, prod, logMode)
+	database, err = db.New(pgURI, pgReadUri, logMode)
 	if err != nil {
 		logger.Fatal(err)
 	}
+	go database.RestoreConnectionWorker(ctx, time.Second*10, pgURI)
 
-	go mq.FatalWorker(time.Second * 10)
```
EnoRage, line 74: can we have fatal workers at the same place?

EnoRage (approved):
Let's try to run it

PR #1221, opened 2020-09-09
[COINS] Change BSC ID

EnoRage:
https://github.com/trustwallet/wallet-core/pull/1101/files

Since it is breaking - we will not support old ID as well


vikmeup:
Can we add as separate id for now? We will remove later? 

And use smartchain as a handle, and symbol just BNB, no need for preferred symbol

PR #1238, opened 2020-10-07
[Elrond] Change symbol to EGLD

andreibancioiu:
Change symbol to EGLD.


vikmeup (approved):

coin/coins.yml
```diff
@@ -390,7 +390,7 @@
   sampleAddress: 'NEAR6Y66fCzeKqWiwxoPox5oGeDN9VhNCu7CEQ9M86iniqoN9vg2X'
 
 - id: 508
-  symbol: ERD
+  symbol: EGLD
```
vikmeup, line 393: `eGLD` is used in the wallet core implementation, should be use `eGLD` for consistency?

andreibancioiu: Actually, we wanted to use `eGLD` here as well, but then we would have broken the _enum_ in `coins.go` (more precisely, the lowercase would have changed the access modifier to private). In the user interface, we'd prefer to have `eGLD` displayed (instead of the full-uppercase version).

vikmeup: let's keep as is, it will be deployed `eGLD` in UI, as it's not using blockatlas to popular symbol for the coin.

PR #1285, opened 2020-11-26
Update elrond integration

miiu96:
Update URL that is used for Elrond API's:

- https://gateway.elrond.com is used instead of https://api.elrond.com

Changed API route that is used to fetch blocks. 

- `URL/hyberblock/by-nonce/blockNonce` is used instead of `URL/metchainShardID/blockNonce`


platform/elrond/model.go
```diff
@@ -47,6 +49,19 @@ type Transaction struct {
 	Timestamp time.Duration `json:"timestamp"`
 	Status    string        `json:"status"`
 	Fee       string        `json:"fee"`
+	GasPrice  uint64        `json:"gasPrice,omitempty"`
```
vikmeup, line 52: is it possible just to use `Fee` ? if gas price and gas limit is optional

platform/elrond/model.go
```diff
@@ -38,6 +39,7 @@ type TransactionsPage struct {
 }
 
 type Transaction struct {
+	Type      string        `json:"type"`
```
vikmeup, line 42: `type` is not used, can we remove it?

config.yml
```diff
@@ -220,7 +220,7 @@ near:
   api: https://staging-rpc.nearprotocol.com
 
 elrond:
-  api: https://api.elrond.com
+  api: https://gateway.elrond.com
```
vikmeup, line 223: is it only block related only? there is also transactions list and status endpoint are used?

platform/elrond/model.go
miiu96: For the moment the fields GasPrice and GasLimit cannot be removed because we use these fields to do an estimation of the transaction fee.

platform/elrond/model.go
miiu96: Field removed.

config.yml
miiu96: Is not a problem if the transactions list is returned by this endpoint. This endpoint is similar with https://api.elrond.com but have some extra API's.

PR #1293, opened 2020-12-03
Improve parser

EnoRage:


config.yml
```diff
@@ -6,16 +6,14 @@ gin:
 # If all - run all platforms in one binary. You can pick specific coin handle to run binary only with specific coin
 # Example: ethereum
 # You can see all the coin handles at coins/coins.yml file
-platform: [ all ]
+platform: [ bitcoin ]
```
vikmeup, line 9:
```suggestion
platform: [ all ]
```

services/parser/parser.go
```diff
@@ -42,7 +42,12 @@ type (
 	}
 )
 
-const MinTxsBatchLimit = 500
+const (
```
vikmeup, line 45: can we move this to config?

services/parser/parser.go
```diff
@@ -67,6 +72,18 @@ func GetInterval(value int, minInterval, maxInterval time.Duration) time.Duratio
 	return time.Duration(pMax)
 }
 
+func GetBlocksByRound(blockTime int, parsingBlocksInterval time.Duration) int64 {
```
vikmeup, line 75: Test?

services/parser/parser.go
```diff
@@ -115,10 +132,12 @@ func GetBlocksIntervalToFetch(params Params, ctx context.Context) (int64, int64,
 		lastParsedBlock = currentBlock - int64(params.BacklogCount)
 	}
 
-	if currentBlock-lastParsedBlock > params.MaxBacklogBlocks {
-		lastParsedBlock = currentBlock - params.MaxBacklogBlocks
+	nextParsedBlock := lastParsedBlock + params.BlocksPerRound
```
vikmeup, line 135: can we test this too?

services/parser/parser.go
EnoRage: any sense? it is not changable for years?

services/parser/parser.go
EnoRage: we use integration tests for parser

services/parser/parser.go
vikmeup: For consistency, most params we specify in config

services/parser/parser.go
vikmeup: Where is the test for this? I see change but there is no tests that will cover this case

services/parser/parser.go
```diff
@@ -67,6 +71,23 @@ func GetInterval(value int, minInterval, maxInterval time.Duration) time.Duratio
...
+	case blockTime >= 4000 && blockTime < 8000:
+		return 18
+	case blockTime >= 8000 && blockTime < 12000:
+		return 12
+	case blockTime >= 12000 && blockTime < 16000:
+		return 10
+	case blockTime >= 16000:
+		return 8
+	default:
+		return DefaultBlocksPerRound
```
vikmeup, line 87: I suggest to remove `GetBlocksByRound` function, because you have control over how often you check for blocks instead. 

Two ways:
check blocks every 1 sec, get 5 blocks
check every 5 sec, get 25 blocks. 

This kinda solves the same problem. so you can control param #1, just define how often you check.

Ideally there is a limit of 10 blocks per blockchain at the time. It could be *2 if behind as an example. 

But `GetBlocksByRound` implementation is overkill as well

tests/integration/observer_test/parser_test.go
```diff
@@ -112,8 +111,8 @@ func setupParser(stopChan chan struct{}) parser.Params {
 		Api:                   getMockedBlockAPI(),
 		ParsingBlocksInterval: pollInterval,
 		BacklogCount:          backlogCount,
-		MaxBacklogBlocks:      int64(maxBatchBlocksAmount),
 		TxBatchLimit:          100,
 		StopChannel:           stopChan,
+		BlocksPerRound: 50,
```
vikmeup, line 116:
```suggestion
		BlocksPerRound: 50,
```

Even for test, probably test 10-20 max. there is no need to fetch more

PR #1297, opened 2020-12-04
Add new mq client. Adapt parser and consumer to new mq client.

madcake:


cmd/consumer/main.go
```diff
@@ -2,19 +2,18 @@ package main
...
+	"time"
+
+	"github.com/trustwallet/blockatlas/services/notifier"
 
 	log "github.com/sirupsen/logrus"
+	"github.com/trustwallet/blockatlas/config"
 	"github.com/trustwallet/blockatlas/db"
 	"github.com/trustwallet/blockatlas/internal"
-	"github.com/trustwallet/blockatlas/mq"
+	"github.com/trustwallet/blockatlas/new_mq"
```
hewigovens, line 16: how about `mqueue`? sounds better than `new_`

hewigovens (approved):
LGTM, probably need to test it a bit before production

vikmeup:
> LGTM, probably need to test it a bit before production

It's work in progress, almost done. We have done the testing in different phases, connect/disconnect mq/postres to see how it reconnect and continues the flow.

cmd/consumer/main.go
vikmeup: It will be just `mq`, `new_mq` will become `mq` as there is no need to keep the old one

PR #1335, opened 2020-12-25
[Filecoin] Add transactions list

### LINKED ISSUE #1246: [Filecoin] Transaction list api

hewigovens:
Fixes #1246


madcake (approved):

config.yml
```diff
@@ -210,6 +210,7 @@ elrond:
 
 filecoin:
   api: https://api.filscan.io:8700/rpc/v1
+  explorer: https://filfox.info/api
```
vikmeup, line 213: Can we drop /api here? And use that in the client instead

vikmeup (approved):

config.yml
hewigovens: removed

PR #1337, opened 2020-12-28
Update issue templates and labels

hewigovens:
Simplify labels, less templates, code owner updates


vikmeup (commented):
Can we remove templates at all? Or just keep one for any issue.

hewigovens:
let's keep just 2: bug report and feature request

PR #1340, opened 2020-12-30
Use address/client/jsonrpc from golibs

hewigovens:


services/notifier/base.go
```diff
@@ -58,3 +58,16 @@ func RunNotifier(database *db.Instance, delivery amqp.Delivery) error {
 
 	return nil
 }
+
+func UnprefixedAddress(address string) (string, uint, bool) {
```
vikmeup, line 62: can you add test for this one just in case?

services/notifier/base.go
```diff
@@ -58,3 +58,16 @@ func RunNotifier(database *db.Instance, delivery amqp.Delivery) error {
...
+	result := strings.Split(address, "_")
+	if len(result) != 2 {
+		return "", 0, false
+	}
+	id, err := strconv.Atoi(result[0])
+	if err != nil {
+		return "", 0, false
+	}
+	return result[1], uint(id), true
+
```
vikmeup, line 72: extra space

vikmeup (approved):

services/notifier/base.go
hewigovens: done

PR #1343, opened 2021-01-03
[FIO] Ignore tx sent to fio.treasury

hewigovens:
Remove hardcode fee memo check


vikmeup (approved):

platform/fio/transaction.go
```diff
@@ -54,8 +55,8 @@ func (p *Platform) Normalize(action *Action, account string) (blockatlas.Tx, err
 			if json.Unmarshal(dataJSON, &actionData) != nil {
 				return blockatlas.Tx{}, errors.New("Unparseable Data field")
 			}
-			if actionData.Memo == "FIO API fees. Thank you." {
-				return blockatlas.Tx{}, errors.New("Skip meaningless hardcoded fee action")
+			if actionData.To == "fio.treasury" {
```
vikmeup, line 58: Maybe add to const?

vikmeup (approved):

platform/fio/transaction.go
hewigovens: done

PR #1350, opened 2021-01-07
Consolidate bitcoin client with blockbook

vikmeup:


platform/bitcoin/base.go
```diff
@@ -2,18 +2,19 @@ package bitcoin
 
 import (
 	"github.com/trustwallet/blockatlas/internal"
+	"github.com/trustwallet/blockatlas/platform/ethereum/blockbook"
```
hewigovens, line 5: move `blockbook` out of ethereum folder?

vikmeup: how about move into `bitcoin` folder?

hewigovens: `pkg/blockbook`?

hewigovens (approved):

PR #1373, opened 2021-01-18
Bug fix elrond integration 

miiu96:
Ignore smart contract results with a negative value
These smart contract results are generated by the staking smart contract --> should be ignored


platform/elrond/transaction.go
```diff
@@ -25,6 +25,10 @@ func NormalizeTxs(srcTxs []Transaction, address string, block Block) (txs []type
 
 // NormalizeTx converts an Elrond transaction into the generic model
 func NormalizeTx(srcTx Transaction, address string, block Block) (tx types.Tx, ok bool) {
+	if srcTx.HasNegativeValue() {
```
vikmeup, line 28: you mentioned that this only happens for smart contract transactions, maybe better would be to avoid processing smart contract transactions and only allow transfers?

hewigovens:
can you do a rebase? too many merge from master commits

PR #1386, opened 2021-01-25
[Collectible] check coins in collectible api

hewigovens:
also update golibs/coin package


config.yml
```diff
@@ -216,10 +216,10 @@ filecoin:
   api: https://api.filscan.io:8700/rpc/v1
   explorer: https://filfox.info
 
-# smartchain:
-#   api: <blockbook api>
-#   rpc: https://bsc-dataseed.nariox.org
-#   collections_api: https://nftview.bounce.finance
+smartchain:
+  api: <blockbook api>
+  rpc: https://bsc-dataseed.binance.org
```
vikmeup, line 221: `smartchain.rpc` is this still required url? I don't think we ever use it in the blockatlas, same for ethereum

platform/harmony/transaction.go
```diff
@@ -64,7 +64,7 @@ func NormalizeTx(trx *Transaction) (tx types.Tx, b bool, err error) {
 
 	return types.Tx{
 		ID:       trx.Hash,
-		Coin:     coin.ONE,
+		Coin:     coin.HARMONY,
```
vikmeup, line 67:
```suggestion
		Coin:     coin.HARMONY,
```

maybe we use `coin.Harmony().ID` for consistency?

vikmeup (approved):

platform/harmony/transaction.go
hewigovens: I'd like keep direct accessing `coin.HARMONY`, unlike `name`, `symbol` it's defined in the map, less call

config.yml
hewigovens: ok, let me remove it

hewigovens: done

PR #1392, opened 2021-01-28
[Collectible] Return category

hewigovens:


platform/ethereum/bounce/collection.go
```diff
@@ -83,11 +83,16 @@ func (c *Client) NormalizeCollectibles(collectibles []Collectible, coinIndex uin
 		return nil, err
 	}
 	for _, c := range collectibles {
+		category := c.ContractName
```
vikmeup, line 86: add a test?

vikmeup (approved):

platform/ethereum/bounce/collection.go
hewigovens: done

PR #1413, opened 2021-02-16
[Solana] support block parsing

### LINKED ISSUE #1361: [Solana] block parsing
RPC:

1. `getConfirmedBlock`
2. `getEpochInfo`

hewigovens:
Fixes #1361 

Changes:
- Add block parsing
- Code cleanup, staking api will be implemented in client


platform/solana/transaction.go
```diff
@@ -2,43 +2,66 @@ package solana
...
 	results := make(types.TxPage, 0)
-	txs, err := p.client.GetTransactions(address)
+	txs, err := p.client.GetTransactionsByAddress(address)
 	if err != nil {
 		return results, err
 	}
 	for _, tx := range txs {
-		if normalized, err := p.NormalizeTx(tx, address); err == nil {
+		timestamp := tx.BlockTime
+		if timestamp == 0 {
```
vikmeup, line 29: in which cases block time is 0?

platform/solana/transaction.go
```diff
@@ -2,43 +2,66 @@ package solana
...
+			timestamp = EstimateTimestamp(tx.Slot)
```
vikmeup, line 30: can we remove `EstimateTimestamp` ? it's not accurate

platform/solana/block.go
```diff
@@ -0,0 +1,25 @@
...
+	return p.client.GetLasteBlock()
+}
+
+func (p *Platform) GetBlockByNumber(num int64) (*types.Block, error) {
+	block, err := p.client.GetTransactionsInBlock(num)
+	if err != nil {
+		return nil, err
+	}
+
+	page := make(types.TxPage, 0)
```
vikmeup, line 15: `types.TxPage` can we rename this to `types.Txs` ?

platform/solana/transaction.go
hewigovens: this `field` is newly added, old data might not have it

vikmeup: let's skip such transactions instead

platform/solana/transaction.go
hewigovens: done

platform/solana/transaction.go
```diff
@@ -2,43 +2,65 @@ package solana
...
 	if err != nil {
 		return results, err
 	}
 	for _, tx := range txs {
-		if normalized, err := p.NormalizeTx(tx, address); err == nil {
+		if tx.BlockTime == 0 {
+			continue
+		}
+		if normalized, err := p.NormalizeTx(tx, tx.Slot, tx.BlockTime); err == nil {
+			if address == normalized.From {
```
vikmeup, line 32: can we you add extension method to transaction `GetDirection() -> Direction` ?  to avoid having this logic here

vikmeup (approved):
Couple of suggestions, otherwise lgtm

platform/solana/transaction.go
hewigovens: thanks, there is an existing method to get direction

PR #1418, opened 2021-02-18
[Solana] ignore sending empty requests

hewigovens:
Fixes 
```json
{
  "error": {
    "message": "unexpected end of JSON input"
  }
}

```


vikmeup (approved):
we need to add silence exposing internal errors too, just return default error

PR #1421, opened 2021-02-19
[OASIS] Add Oasis platform

Anonyma:
BASE INTEGRATION `platform/oasis`
   - [x] `base.go`: Init() & Coin() config
   - [x] `model.go`: Platform-specific models	   
   - [x] `client.go`: Platform getter methods (API)
   - [x] `transaction.go`: `NormalizeTx` method to match Atlas model + `GetTxsByAddress` method
   - [x] `block.go`: `CurrentBlockNumber` & `GetBlockByNumber` methods
 
CONFIG
 - [x] `config.yml` + `config/configuration.go`: Add API URL

REST API
- [x] Endpoint to query a list of transactions related to an address
- [x] Endpoint to query a list of transactions from a specific block height
- [x] Endpoint to query latest block height


platform/oasis/base.go
```diff
@@ -0,0 +1,22 @@
...
+
+func Init(api string) *Platform {
+	p := &Platform{
+		client: Client{client.InitClient(api, middleware.SentryErrorHandler)},
+	}
+	return p
+}
+
+func (p *Platform) Coin() coin.Coin {
+	return coin.Coins[coin.OASIS]
```
hewigovens, line 21:
```suggestion
	return coin.Oasis()
```

platform/oasis/transaction.go
```diff
@@ -0,0 +1,54 @@
...
+		Coin:     coin.OASIS,
+		From:     srcTx.From,
+		To:       srcTx.To,
+		Fee:      types.Amount(srcTx.Fee),
+		Date:     srcTx.Date,
+		Block:    srcTx.Block,
+		Status:   status,
+		Error:    srcTx.ErrorMsg,
+		Sequence: srcTx.Sequence,
+		Meta: types.Transfer{
```
hewigovens, line 46: can you ignore non transfer transactions for now?

config.yml
```diff
@@ -213,6 +213,9 @@ elrond:
 filecoin:
   api: https://api.filscan.io:8700/rpc/v1
   explorer: https://filfox.info
+  
+oasis:
+  api: https://oasis.atlas.zondax.ch
```
hewigovens, line 218: can we use this endpoint in production?

platform/oasis/transaction.go
```diff
@@ -0,0 +1,54 @@
+package oasis
+
+import (
+	"github.com/trustwallet/golibs/coin"
+	"github.com/trustwallet/golibs/types"
+)
+
+func (p *Platform) GetTxsByAddress(address string) (types.TxPage, error) {
```
hewigovens, line 8: please rebase master, `GetTxsByAddress` should return `types.Txs` now
```suggestion
func (p *Platform) GetTxsByAddress(address string) (types.Txs, error) {
```

platform/oasis/transaction.go
```diff
@@ -0,0 +1,54 @@
...
+	txs, err := p.client.GetTrxOfAddress(address)
+	if err != nil {
+		return nil, err
+	}
+
+	return NormalizeTxs(*txs), nil
+}
+
+func NormalizeTxs(srcTxs []Transaction) types.TxPage {
```
hewigovens, line 17:
```suggestion
func NormalizeTxs(srcTxs []Transaction) types.Txs {
```

hewigovens (commented):
do you have sample address to test txs api?

platform/oasis/transaction.go
hewigovens: @Anonyma can you fix this?

Anonyma: @hewigovens yes, only transfer transactions are returned from the API, if that is what you mean

config.yml
Anonyma: Please use it for testing for now

> do you have sample address to test txs api?

Here is an example address: `oasis1qpqu7cfadhayekanmrtk8t3v4cx9483gmqu7h6gn`
https://www.oasisscan.com/accounts/detail/oasis1qzt0cma7p9yyuqdz56kkg0t7g9al96glm50aufmy

platform/oasis/block.go
```diff
@@ -0,0 +1,25 @@
...
+}
+
+func (p *Platform) GetBlockByNumber(num int64) (*types.Block, error) {
+	if srcBlock, err := p.client.GetBlockByNumber(num); err == nil {
+		txs := NormalizeTxs(*srcBlock)
+		return &types.Block{
+			Number: num,
+			Txs:    txs,
+		}, nil
+	} else {
```
vikmeup, line 22: can you remove else and do `return nil, err`

platform/oasis/transaction_test.go
```diff
@@ -0,0 +1,95 @@
...
+func TestNormalizeTx(t *testing.T) {
+	tests := []struct {
+		name     string
+		filename string
+		wantTx   types.Tx
+	}{
+		{
+			name:     "Test normalize successful transaction without fee",
+			filename: "tx_without_fee.json",
+			wantTx: types.Tx{
```
vikmeup, line 20:
```
Memo: ""
Error: "",
```

you can probably skip it as it's empty by default

vikmeup (approved):

PR #1446, opened 2021-04-22
Add user agent header

iulianpascalau:
Added user agent when doing a request on the trezor servers since, apparently those servers require that header. Without this header "http error 403 - Forbidden" occurs.


Amenocy:
please merge this one :)

PR #1478, opened 2022-03-15
upgraded some packages & go mod tidy

vorotech:


hewigovens (approved):
Sorry, I think we should archive this repo

